PR #1, opened 2021-11-24
Add baseline simulation functionality, from prototype developed Q3 2021

stephanie-chan-cnz:
This PR adds baseline functionality to run a simulation using a local .csv dataset of household attributes.


Pipfile
```diff
@@ -4,6 +4,10 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
+pandas = "*"
+tqdm = "*"
+numpy = "*"
+typing-extensions = "*"
```
tomwphillips, line 10: Why have you added this? The [description on PyPI](https://pypi.org/project/typing-extensions/) says it's a backport of `typing`, but we're using Python 3.9.

simulation/settings.py
```diff
@@ -0,0 +1,24 @@
...
+import pandas as pd
+
+
+def parse_args(args=None):
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--household-distribution", type=pd.read_csv)
+    return parser.parse_args(args)
+
+
+args = parse_args()
```
tomwphillips, line 13: If you have this here then as soon as anything is imported from this module, `parse_args` will be excuted and try to parse arguments. Instead `__main__` should import `get_args` and call it there.

I know a settings file has been mentioned a lot lately. I'm not convinced it's a panacea.

Genuine constants should be constants in the appropriate Python module.

Things like `NUM_STEPS`, `START_DATETIME` and `NUM_HOUSEHOLDS` are simulation parametes, so I think they should go in `parse_args` with sensible defaults, so when the user runs the simulation they can be changed on the command line.

If everyone really wants to specify things in a file, then we could have an argument for a file path and load settings from that instead of the command line.

simulation/model.py
```diff
@@ -0,0 +1,102 @@
...
+
+    @property
+    def current_heating_system_count(self) -> Dict[HeatingSystem, int]:
+        return Counter(
+            [
+                agent.heating_system
+                for agent in self.space.agents
+                if isinstance(agent, Household)
+            ]
+        )
```
tomwphillips, line 43: I don't recognise these two properties from before. There's no tests either. Is it new?

I think these things would be better suited for model collectors.

We would know the number of agents anyway because agents aren't removed in the simulation.

tomwphillips (changes requested):

simulation/agents.py
```diff
@@ -0,0 +1,74 @@
...
+    HEATING_SYSTEM_FUEL,
+    HEATING_SYSTEM_LIFETIME_YEARS,
+    BuiltForm,
+    Epc,
+    HeatingFuel,
+    HeatingSystem,
+    OccupantType,
+    PropertyType,
+)
+from simulation.settings import HEAT_PUMP_AWARENESS
```
RyanJenkinson, line 16: I think the `settings` file needs a little bit of thought: how would we run multiple simulations on google cloud, for example? We would want multiple `settings` files, but the current implementation where we import from `simulation.settings` makes that hard to do.

I would be open to a `settings.yaml` (or similar) file that is passed through in `parse_args()`, but not sure about the best way to import those settings into the `agents.py` here. One thought would be to have a section in the yaml/toml file called "agents", and every argument there is passed into the `Household.__init__()` (e.g `heat_pump_aware_probability`. This would allow us to also easily test for (non) heat pump aware agents, since we can set that value to 0 or 1 respectively in the tests.

simulation/agents.py
```diff
@@ -0,0 +1,74 @@
...
+    ):
+        # Property / tenure attributes
+        self.location = location
+        self.property_type = property_type
+        self.occupant_type = occupant_type
+        self.built_form = built_form
+        self.floor_area_sqm = floor_area_sqm
+        self.property_value = property_value
+        self.is_solid_wall = is_solid_wall
+        self.construction_year = self.sample_uniformly_from_interval(
```
RyanJenkinson, line 46: Do we need this staticmethod here, any more? Can we just use `random.randint(...)` since we want a full year and not a float, anyway? If so, we can simplify this a bit more and remove the `staticmethod` below.
N.B if we do want to estimate a construction datetime, then we should do that instead with a datetime object.

simulation/constants.py
```diff
@@ -0,0 +1,56 @@
...
+    MID_TERRACE = 0
+    SEMI_DETACHED = 1
+    DETACHED = 2
+    END_TERRACE = 3
+
+
+class OccupantType(enum.Enum):
+    OWNER_OCCUPIER = 0
+    RENTER_PRIVATE = 1
+    RENTER_SOCIAL = 1
```
RyanJenkinson, line 21:
```suggestion
    RENTER_SOCIAL = 2
```
Did you want these all to be different or actually the 0 or 1 so that when we convert to `OccupantType.value` we essentially get `is_renter` ?
To be honest, I quite like the latter approach - just wanted to check it was intentional here?

simulation/constants.py
```diff
@@ -0,0 +1,56 @@
...
+    F = 5
+    G = 6
+
+
+class HeatingSystem(enum.Enum):
+    BOILER_GAS = 0
+    BOILER_OIL = 1
+    BOILER_ELECTRIC = 2
+    HEAT_PUMP_AIR_SOURCE = 3
+    HEAT_PUMP_GROUND_SOURCE = 4
```
RyanJenkinson, line 39: Note we could play the same game as above e.g 
```suggestion
    BOILER_GAS = 0
    BOILER_OIL = 0
    BOILER_ELECTRIC = 0
    HEAT_PUMP_AIR_SOURCE = 1
    HEAT_PUMP_GROUND_SOURCE = 1
```
so that when we apply `HeatingSystem.value` it turns into `is_heat_pump`. Or is this an abuse of `enum.Enum` and really they should all be different? @tomwphillips ?

simulation/constants.py
```diff
@@ -0,0 +1,56 @@
...
+
+HEATING_SYSTEM_FUEL: Dict[HeatingSystem, HeatingFuel] = {
+    HeatingSystem.BOILER_GAS: HeatingFuel.GAS,
+    HeatingSystem.BOILER_OIL: HeatingFuel.OIL,
+    HeatingSystem.BOILER_ELECTRIC: HeatingFuel.ELECTRICITY,
+    HeatingSystem.HEAT_PUMP_AIR_SOURCE: HeatingFuel.ELECTRICITY,
+    HeatingSystem.HEAT_PUMP_GROUND_SOURCE: HeatingFuel.ELECTRICITY,
+}
+
+HEATING_SYSTEM_LIFETIME_YEARS = 15
```
RyanJenkinson, line 56: Should this live here? Or in the settings file? Is it something we would want to potentially vary?

simulation/model.py
```diff
@@ -0,0 +1,102 @@
...
+    ].apply(parse_string_interval)
+
+    households = household_distribution.sample(num_households, replace=True)
+    for household in households.itertuples():
+        yield Household(
+            location=household.location,
+            property_value=household.property_value,
+            floor_area_sqm=household.floor_area_sqm,
+            off_gas_grid=household.off_gas_grid,
+            construction_year_band=household.construction_year_band,
```
RyanJenkinson, line 71: I think we should change the construction year band to either:
1. A single int value `construction_year: int`, of the year it was constructed, and do the preprocessing from the string to that value at this stage
2. A `enum.Enum` for `ConstructionYearBand` which has all the property year archetypes we are using in there. We should map the string onto that enum at this stage, too. Then `construction_year_band: ConstructionYearBand` and we can work with it a bit easier.

I would be in favour of 2.

RyanJenkinson: Note: if we did 2 we can remove all the `pd.Interval` stuff we have, and preprocessing functions etc.

RyanJenkinson (changes requested):
Just some other comments / thoughts that build on Toms above. Happy to discuss!

simulation/agents.py
tomwphillips: >I think the settings file needs a little bit of thought: how would we run multiple simulations on google cloud, for example? We would want multiple settings files, but the current implementation where we import from simulation.settings makes that hard to do.

This is exactly why a Python file of parameters doesn't work. 

For now I think focus on we want to do and I can make it happen. I think we want to be able to:

* run a simulation with some sensible defaults
* run a simulation with some different parameters
* run lots of simulations with different parameters (a bit like gridsearch for hyperparameter optimisation)

simulation/constants.py
tomwphillips: I think that'll be easy to miss. If we don't need the private/social distinction remove it otherwise make them different values.

simulation/constants.py
tomwphillips: Like I said above it's a bit unexpected IMO.

Pipfile
stephanie-chan-cnz: Hm, docker build failed without it, raising an error about `typing-extensions` missingness. I'll remove and see what happens again.

simulation/model.py
stephanie-chan-cnz: Not new, as per previous repo. Agreed, we can remove it.

simulation/agents.py
stephanie-chan-cnz: Let's discuss what we think are our requirements together (maybe with @soolinglim) and come back to the points around using settings files.

simulation/constants.py
stephanie-chan-cnz: This was a typo - I intended them to represent different values. While the distinction between renters does not alter agent behaviour, but is potentially a field we would want to segment simulation outputs by, so I'd prefer not to lose this detail.

simulation/constants.py
stephanie-chan-cnz: Maybe, but I wouldn't put it on say the top 25 things we'd want to vary, as it's far less open to debate than many other parameters.

simulation/model.py
stephanie-chan-cnz: I don't mind - I feel that sampling uniformly from the EPC construction year bands results is unlikely to be the best approach anyway (there were probably some property booms within these sub-periods, not uniform in count). It would remove any ability to cap heating system age by construction year though (which I removed for this first PR, and I think is fine to continue to omit).

simulation/__main__.py
```diff
@@ -0,0 +1,52 @@
...
+history = create_and_run_simulation(
+    args.start_datetime,
+    args.step_interval,
+    args.num_steps,
+    args.num_households,
+    args.household_distribution,
+    args.heat_pump_awareness,
+)
+
+write_jsonlines(history, args.history_filename)
```
tomwphillips, line 52:
```suggestion
if __name__ == "__main__":
    args = parse_args()

    history = create_and_run_simulation(
        args.start_datetime,
        args.step_interval,
        args.num_steps,
        args.num_households,
        args.household_distribution,
        args.heat_pump_awareness,
    )

    write_jsonlines(history, args.history_filename)
```

tomwphillips (approved):

RyanJenkinson (approved):

PR #2, opened 2021-11-30
Add household properties associated with renovation budget

stephanie-chan-cnz:
- Adds constants representing A) the distribution of GB property values, and B) the distribution of GB renovation budgets
- Household property `wealth_percentile` calculates the percentile of `household.property_value` within distribution A)
- Household property `renovation_budget` calculates the value of at `wealth_percentile` within distribution B), multiplied by a 0.1 factor to represent a proportion of "overall renovation spend" a household might dedicate to heating and energy efficiency measures


simulation/tests/test_agents.py
```diff
@@ -71,3 +73,22 @@ def test_create_household(self) -> None:
...
+        assert (
+            low_property_value_household.renovation_budget
+            < medium_property_value_household.renovation_budget
+            < high_property_value_household.renovation_budget
+        )
+
+    def test_renovation_budget_greater_than_or_equal_to_zero_and_less_than_total_property_value(
+        self,
+    ) -> None:
+        household = household_factory(property_value=random.randint(0, 5_000_000))
```
RyanJenkinson, line 91: Will a house ever have 0 value? We could change this test to select a `randint(1, 5_000_000)` and then just `assert 0 < household.renovation_budget < household.property_value`.

Also just double checking the behaviour here: we are assuming that _all_ houses will have a renovation budget (equal to, currently, 10% of their property value) regardless of any other attributes, for now.

RyanJenkinson (changes requested):
Just some really minor changes, but looks great!

Another thought I had, potentially for a future PR, is that we create a `simulation/distributions.py` with a class `Weibull(alpha, beta)` which has methods `percentile_from_value` and `value_from_percentile` and potentially other helpful properties like the `mean`. I don't propose we do it as a part of this PR, but one to think about as we start using more distributions, if it makes sense to refactor it this way.

simulation/agents.py
```diff
@@ -60,5 +65,62 @@ def __init__(
...
+
+    @property
+    def renovation_budget(self) -> float:
+        # An amount a house may set aside for work related to home heating and energy efficiency
+        # Expressed as a proportion of their total renovation budget (10%)
+
+        """
+        :return: the estimated renovation budget of a household, derived from the value of a distribution of GB
+        renovation budgets at a household's wealth percentile
+        """
```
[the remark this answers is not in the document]
stephanie-chan-cnz: Let's just do comments for now I think then, keep the code leaner.

simulation/tests/test_agents.py
stephanie-chan-cnz: I think DQ of the source data should primarily be covered within data warehouse - e.g. non-zero property values - rather than kind of mixed into this test.

Not quite - all houses have a renovation budget, which is drawn from a distribution of GB renovation budgets at the same percentile as their relative property value. So this deliberately doesn't need to correspond to 10% of their property value (in practice ranges between 0-25% for most households). 10% of this overall budget is allocated by a house to "heating systems and energy efficiency measures" - the part of the budget relevant to the future heating system decision logic.

RyanJenkinson (approved):

PR #3, opened 2021-11-30
Add functionality for households to hold renovations

stephanie-chan-cnz:
Adds functionality for households to evaluate whether to hold a renovation, and decide the scope of the renovation.

### DIFF
--- a/simulation/__main__.py
+++ b/simulation/__main__.py
@@ -34,6 +34,7 @@ def parse_args(args=None):
     parser.add_argument("--history-filename", default="history.jsonl")
     parser.add_argument("--household-distribution", type=pd.read_csv)
     parser.add_argument("--heat-pump-awareness", type=float, default=0.4)
+    parser.add_argument("--annual-renovation-rate", type=float, default=0.05)
 
     return parser.parse_args(args)
 
@@ -48,6 +49,7 @@ if __name__ == "__main__":
         args.num_households,
         args.household_distribution,
         args.heat_pump_awareness,
+        args.annual_renovation_rate,
     )
 
     write_jsonlines(history, args.history_filename)
--- a/simulation/agents.py
+++ b/simulation/agents.py
@@ -1,6 +1,7 @@
 import datetime
 import math
 import random
+from typing import Dict
 
 from abm import Agent
 from simulation.constants import (
@@ -113,7 +114,19 @@ class Household(Agent):
 
         self.is_renovating = True if random.random() < proba_renovate else False
 
+    def decide_renovation_scope(self) -> Dict[str, bool]:
+
+        PROBA_HEATING_SYSTEM_UPDATE = 0.18
+        PROBA_INSULATION_UPDATE = 0.33
+
+        return {
+            "HEATING_SYSTEM": True
+            if random.random() < PROBA_HEATING_SYSTEM_UPDATE
+            else False,
+            "INSULATION": True if random.random() < PROBA_INSULATION_UPDATE else False,
+        }
+
     def step(self, model):
         self.evaluate_renovation(model)
         if self.is_renovating:
-            pass
+            renovation_scope = self.decide_renovation_scope()
--- a/simulation/agents.py
+++ b/simulation/agents.py
@@ -116,6 +116,8 @@ class Household(Agent):
 
     def decide_renovation_scope(self) -> Dict[str, bool]:
 
+        # Derived from the VERD Project, 2012-2013. UK Data Service. SN: 7773, http://doi.org/10.5255/UKDA-SN-7773-1
+        # Based upon the choices of houses in 'Stage 3' - finalising or actively renovating
         PROBA_HEATING_SYSTEM_UPDATE = 0.18
         PROBA_INSULATION_UPDATE = 0.33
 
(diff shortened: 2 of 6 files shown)


simulation/__main__.py
```diff
@@ -34,6 +34,7 @@ def convert_to_timedelta(minutes_string):
     parser.add_argument("--history-filename", default="history.jsonl")
     parser.add_argument("--household-distribution", type=pd.read_csv)
     parser.add_argument("--heat-pump-awareness", type=float, default=0.4)
+    parser.add_argument("--annual-renovation-rate", type=float, default=0.05)
```
RyanJenkinson, line 37: Think we might need a `help=...` here with a bit more description about the rate. Might not be in this PR, but can do another one adding help statements to each of the arguments as I think it would be useful.

e.g the `rate` here refers to "per year" (from reading the code below)

simulation/agents.py
```diff
@@ -102,5 +107,28 @@ def renovation_budget(self) -> float:
             self.wealth_percentile,
         )
 
+    def evaluate_renovation(self, model) -> None:
+
+        step_interval_years = model.step_interval / datetime.timedelta(days=365)
+        proba_renovate = model.annual_renovation_rate * step_interval_years
+
+        self.is_renovating = True if random.random() < proba_renovate else False
```
RyanJenkinson, line 115: We will have a lot of the `True if random.random() < proba_renovate else False` statements. Note that this is just the same as:
```suggestion
        self.is_renovating = (random.random() < proba_renovate)
```
and we should replace all instances in the code with that. 

I'm wondering if there is a simpler `@staticmethod` we can add to the agent since we will use this a lot. I'm struggling to think of a good, concise name, though, but something like `true_with_probability(p: float)`

what do you think of:
```
@staticmethod
def true_with_probability(p: float) -> bool:
    return random.random() < p
```

simulation/agents.py
```diff
@@ -102,5 +107,28 @@ def renovation_budget(self) -> float:
...
+        # Derived from the VERD Project, 2012-2013. UK Data Service. SN: 7773, http://doi.org/10.5255/UKDA-SN-7773-1
+        # Based upon the choices of houses in 'Stage 3' - finalising or actively renovating
+        PROBA_HEATING_SYSTEM_UPDATE = 0.18
+        PROBA_INSULATION_UPDATE = 0.33
+
+        return {
+            "HEATING_SYSTEM": True
+            if random.random() < PROBA_HEATING_SYSTEM_UPDATE
+            else False,
+            "INSULATION": True if random.random() < PROBA_INSULATION_UPDATE else False,
```
RyanJenkinson, line 128: See comment above

simulation/agents.py
```diff
@@ -102,5 +107,28 @@ def renovation_budget(self) -> float:
...
         )
 
+    def evaluate_renovation(self, model) -> None:
+
+        step_interval_years = model.step_interval / datetime.timedelta(days=365)
+        proba_renovate = model.annual_renovation_rate * step_interval_years
+
+        self.is_renovating = True if random.random() < proba_renovate else False
+
+    def decide_renovation_scope(self) -> Dict[str, bool]:
```
RyanJenkinson, line 117: Does this function need to return a dictionary or can it just set internal variables e.g
```
self.renovate_heating_system = True / False
self.renovate_insulation = True / False
```

simulation/tests/test_agents.py
```diff
@@ -92,3 +95,18 @@ def test_renovation_budget_greater_than_or_equal_to_zero_and_less_than_total_pro
...
         assert household.renovation_budget < household.property_value
+
+    def test_household_is_renovating_state_updates(
+        self,
+    ) -> None:
+
+        model = CnzAgentBasedModel(
+            start_datetime=datetime.datetime.now(),
+            step_interval=datetime.timedelta(days=365),
+            annual_renovation_rate=1.0,
```
RyanJenkinson, line 106: Nice way of testing

simulation/tests/test_collectors.py
```diff
@@ -4,9 +4,20 @@
 from simulation.model import CnzAgentBasedModel
 
 
+def model_factory(**model_attributes):
```
RyanJenkinson, line 7: I like the idea of a model factory, and agree we should use it more generally across the tests.

RyanJenkinson (changes requested):
Just some minor changes to cut down the code a little. Generally looks good, though!

simulation/agents.py
RyanJenkinson: I'm also happy for this not to be a staticmethod, but just to be a function defined above the agents itself. That way, we will save a lot of `self.true_with_probability` stuff, and I think this method is general enough it can be defined above.

simulation/agents.py
stephanie-chan-cnz: Could do, what's the advantage of doing so?

simulation/agents.py
```diff
@@ -85,6 +84,10 @@ def get_weibull_value_from_percentile(
         epsilon = 0.0000001
         return beta * (-math.log(1 + epsilon - percentile)) ** (1 / alpha)
 
+    @staticmethod
+    def evaluate_trial(p: float) -> bool:
```
RyanJenkinson, line 88: I do get the name here, but I think that `true_with_probability(p: float)` is easier to read?

RyanJenkinson (approved):
Looks great! Also happy to rename `true_with_probability` to `bernoulli_trial` if you prefer.

PR #4, opened 2021-12-01
Add household methods to get costs for insulation elements

stephanie-chan-cnz:
- Adds method to identify "upgradable" insulation elements. An insulation element refers to the household's roof, windows, or walls. It's upgradable if the energy efficiency is less than 5 (the maximum score from the EPC certificate).
- Adds method to get a cost estimate per element. This is segmented as per the source BEIS report, as different property segments can materially influence costs (e.g. the cost to add wall insulation to a mid-terrace house is very different to a detached house).

### DIFF
--- a/simulation/agents.py
+++ b/simulation/agents.py
@@ -1,7 +1,9 @@
 import datetime
 import math
 import random
-from typing import Set
+from typing import Dict, Set
+
+import pandas as pd
 
 from abm import Agent
 from simulation.constants import (
@@ -17,9 +19,20 @@ from simulation.constants import (
     Epc,
     HeatingFuel,
     HeatingSystem,
+    InsulationSegment,
     OccupantType,
     PropertyType,
 )
+from simulation.costs import (
+    CAVITY_WALL_INSULATION_COST,
+    DOUBLE_GLAZING_UPVC_COST,
+    INTERNAL_WALL_INSULATION_COST,
+    LOFT_INSULATION_JOISTS_COST,
+)
+
+
+def sample_interval_uniformly(interval: pd.Interval) -> float:
+    return random.randint(interval.left, interval.right)
 
 
 class Household(Agent):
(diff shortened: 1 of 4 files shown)


simulation/agents.py
```diff
@@ -129,7 +186,52 @@ def decide_renovation_scope(self) -> None:
...
+        )
+
+        MAX_ENERGY_EFFICIENCY_SCORE = 5
+        return set(
+            [
+                measure
+                for measure, grade in measures_and_grades
+                if grade < MAX_ENERGY_EFFICIENCY_SCORE
+            ]
+        )
```
tomwphillips, line 207: Minor syntax suggestion: like list comprehnsions, you can do set comprehensions, e.g. `{measure for measure, grade in measures_and_grades if grade < MAX_ENERGY_EFFICIENCY_SCORE}`.

simulation/tests/test_agents.py
```diff
@@ -110,3 +113,42 @@ def test_household_is_renovating_state_updates(
...
+    ) -> None:
+
+        household = household_factory(roof_energy_efficiency=np.NaN)
+        assert Element.ROOF not in household.get_upgradable_insulation_elements()
+
+    def test_household_elements_under_max_energy_efficiency_score_are_upgradable(
+        self,
+    ) -> None:
+
+        MAX_ENERGY_EFFICIENCY_SCORE = 5
```
tomwphillips, line 128: Found this a little confusing because you have `MAX_ENERGY_EFFICIENCY_SCORE` here but then there's `MAX_ENERGY_EFFICIENCY_SCORE` in the body of `get_upgradable_insulation_elements` too. What are you trying to get across?

simulation/agents.py
```diff
@@ -110,6 +125,48 @@ def renovation_budget(self) -> float:
             self.wealth_percentile,
         )
 
+    @property
+    def insulation_segment(self) -> InsulationSegment:
```
tomwphillips, line 129: I'm on the fence about asking for tests with this. On the one hand, it's from the BEIS report (can you link it to it?). There's not really any behaviour - it's just mapping.

On the other, it's not a very nice bunch of if statements and I worry about it being unintentionally broken in the future. The behaviours in the get quote method rely on this property working. Maybe a sensible compromise is a test like `test_new_roof_is_cheaper_for_small_bungalow_than_large_detatched`. I'm open to discussion...

simulation/tests/test_agents.py
stephanie-chan-cnz: That 5 is the maximum score in the source dataset. I will update.

simulation/agents.py
stephanie-chan-cnz: Yeah, it's a bit of limbo. It's not ingested from source via the data warehouse, so it has bypassed some of the DQ/validation at that layer. On the other hand, I don't think it really feel it makes sense to devote too much time to such tests as a) I think we'd need to devote disproportionate effort and a large N of permutations to validate here, and b) I think major errors would likely surface out at the post-simulation analysis of our baseline scenario (e.g. if observing that large flats pay less for insulation than small flats).

tomwphillips: Ok. I understand and agree about all the permutations. My preference and recommendation would be to have a couple simple cases, but if you want to leave it for post-simulation analysis, I'm ok with that.

tomwphillips (approved):

PR #7, opened 2021-12-02
Households have heating system breakdowns, and restricted system choices at breakdown

stephanie-chan-cnz:
- Initialise households with `heating_system_install_date`, to replace household `heating_system_age`. This avoids us needing to update heating system age at each timestep.
- Add household method to evaluate if heating is functioning, based on age of heating system and Weibull hazard function parameters
- Modify `household.get_heating_system_options` so heat pumps are not options during a breakdown (due to high industry lead times), unless a household already has a heat pump installed. In that event, the specific type of heat pump already installed (air/ground source) remains an option.

### DIFF
--- a/simulation/agents.py
+++ b/simulation/agents.py
@@ -115,7 +115,7 @@ class Household(Agent):
         return random.random() < p
 
     @staticmethod
-    def weibull_hazard_rate(alpha: float, beta: float, age_years: float):
+    def weibull_hazard_rate(alpha: float, beta: float, age_years: float) -> float:
         # Source: # https://en.wikipedia.org/wiki/Weibull_distribution
 
         """
--- a/simulation/agents.py
+++ b/simulation/agents.py
@@ -54,7 +54,7 @@ class Household(Agent):
         property_type: PropertyType,
         built_form: BuiltForm,
         heating_system: HeatingSystem,
-        heating_system_install_date: datetime.datetime,
+        heating_system_install_date: datetime.date,
         epc: Epc,
         potential_epc: Epc,
         occupant_type: OccupantType,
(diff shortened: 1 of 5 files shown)


simulation/agents.py
```diff
@@ -50,6 +54,7 @@ def __init__(
         property_type: PropertyType,
         built_form: BuiltForm,
         heating_system: HeatingSystem,
+        heating_system_install_date: datetime.datetime,
```
tomwphillips, line 57: Why is `household.heating_system_install_date` a `datetime` and not a `date`?

simulation/tests/test_agents.py
```diff
@@ -241,42 +246,106 @@ def test_households_not_suitable_archetype_are_not_heat_pump_suitable(
...
+        event_trigger,
     ) -> None:
 
         unaware_household = household_factory(is_heat_pump_aware=False)
         model = model_factory()
         assert not HEAT_PUMPS.intersection(
-            unaware_household.get_heating_system_options(model)
+            unaware_household.get_heating_system_options(
+                model, event_trigger=event_trigger
+            )
```
tomwphillips, line 274: I can see it's not something you added in this PR, so don't feel obliged to change it, but I had to think about what's going on here with `assert not` and a set. I think something more explicit like the following is clearer.

```
heating_system_options = unaware_household.get_heating_system_options(model, event_trigger)
assert heating_system_options.intersection(HEAT_PUMPS) == {}
```

simulation/tests/test_agents.py
```diff
@@ -241,42 +246,106 @@ def test_households_not_suitable_archetype_are_not_heat_pump_suitable(
...
-        self,
+        self, event_trigger
     ) -> None:
 
         off_gas_grid_household = household_factory(off_gas_grid=True)
         model = model_factory()
         assert not {HeatingSystem.BOILER_GAS}.intersection(
-            off_gas_grid_household.get_heating_system_options(model)
+            off_gas_grid_household.get_heating_system_options(
+                model, event_trigger=event_trigger
```
tomwphillips, line 286: Again, not obligatory, but no reason for the keyword argument with `event_trigger` in these tests.

tomwphillips (commented):
Just a question about the `datetime` and a couple of other optional things.

simulation/agents.py
stephanie-chan-cnz: True, I'll change it to a date, no need to have a lifetime in microseconds!

tomwphillips (approved):

PR #8, opened 2021-12-02
Households have discount rates linked to wealth percentile

stephanie-chan-cnz:
- Add Weibull parameters approximately corresponding to a distribution of discount rates for households selecting heating system related purchases (Individual Time Preferences and Energy Efficiency - NBER Working Paper No. 20969)
- Add household method to fetch a discount rate value from this distribution, based on wealth percentile. There is an inverse relationship between discount rate and wealth.

### DIFF
--- a/simulation/agents.py
+++ b/simulation/agents.py
@@ -11,6 +11,8 @@ if TYPE_CHECKING:
     from simulation.model import CnzAgentBasedModel
 
 from simulation.constants import (
+    DISCOUNT_RATE_WEIBULL_ALPHA,
+    DISCOUNT_RATE_WEIBULL_BETA,
     GB_PROPERTY_VALUE_WEIBULL_ALPHA,
     GB_PROPERTY_VALUE_WEIBULL_BETA,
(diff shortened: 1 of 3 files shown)


simulation/agents.py
```diff
@@ -118,6 +120,19 @@ def wealth_percentile(self) -> float:
             self.property_value,
         )
 
+    @property
+    def discount_rate(self) -> float:
+
+        return max(
+            1
+            - self.get_weibull_value_from_percentile(
```
RyanJenkinson, line 128: Why do we have `1 -` here and not just the distribution of discount rates as a weibull?

stephanie-chan-cnz: E.g. the 90th wealth percentile corresponds to the 10th percentile of discount rates, because of the studied effect of discount rates being inversely related to income or wealth. The theory is that lower income households do not/cannot value future cashflows as highly as near-term ones, due to more pressing immediate financial needs -- therefore apply higher 'discount' rates to these cashflows.

RyanJenkinson (approved):

PR #9, opened 2021-12-03
Households have heat pump capacity requirements based on floor area

stephanie-chan-cnz:
- Air source and ground source heat pumps are associated with different floor area scaling factors, minimum capacities and maximum capacities.
- Households fetch required capacity for a given type of heat pump on a simple defined relationship: `scale factor * floor area`. This output is capped by the min/max allowed capacity of the given heat pump type.


RyanJenkinson (dismissed):
LGTM!

RyanJenkinson (approved):

PR #10, opened 2021-12-06
Households fetch costs of heating systems (unit, install, and fuel costs)

stephanie-chan-cnz:
- Costs of unit, install and fuel prices are added as constants
- Households fetch unit and install costs (based on required capacity for heat pumps, or coarse property band sizes for boilers). Install costs depend on whether it's a like-for-like replacement, or a switch (which have implied additional costs, like decommissioning redundant system parts).
- Households have an estimated annual kwh heat demand, based on their floor area. Note that we ignore any impact of EPC for the time being. This aligns to some empirical studies which suggest households often increase energy usage after improving efficiency, to further increase comfort levels at the same cost.
- Households evaluate future fuel bills based on some number of "look ahead" years (defined at the model level). For example, while a heating system may last for 20 years, a household may evaluate changes to their costs over a shorter horizon (default: 3 years). Their valuation is also impacted by their own discount rate. Less wealthy households = higher discount rates = future cashflows have a lower present day value.

### DIFF
--- a/simulation/agents.py
+++ b/simulation/agents.py
@@ -32,6 +32,7 @@ from simulation.constants import (
     HeatingSystem,
     InsulationSegment,
     OccupantType,
+    PropertySize,
     PropertyType,
 )
 from simulation.costs import (
@@ -39,6 +40,7 @@ from simulation.costs import (
     DOUBLE_GLAZING_UPVC_COST,
     INTERNAL_WALL_INSULATION_COST,
     LOFT_INSULATION_JOISTS_COST,
+    get_unit_and_install_costs,
 )
 
 HEAT_PUMPS = {HeatingSystem.HEAT_PUMP_AIR_SOURCE, HeatingSystem.HEAT_PUMP_GROUND_SOURCE}
@@ -222,6 +224,20 @@ class Household(Agent):
             else True
         )
 
+    @property
+    def property_size(self) -> PropertySize:
+
+        # Source: England/Wales EPC
+        FLOOR_AREA_SQM_33RD_PERCENTILE = 66
+        FLOOR_AREA_SQM_66TH_PERCENTILE = 89
+
+        if self.floor_area_sqm < FLOOR_AREA_SQM_33RD_PERCENTILE:
+            return PropertySize.SMALL
+        elif self.floor_area_sqm > FLOOR_AREA_SQM_66TH_PERCENTILE:
+            return PropertySize.LARGE
+        else:
+            return PropertySize.MEDIUM
+
     def heating_system_age_years(self, current_date: datetime.date) -> float:
         return (current_date - self.heating_system_install_date).days / 365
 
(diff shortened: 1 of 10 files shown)


simulation/agents.py
```diff
@@ -351,20 +377,29 @@ def update_heating_status(self, model: "CnzAgentBasedModel") -> None:
...
-    def compute_heat_pump_capacity_kw(self, heat_pump_type: HeatingSystem) -> float:
+    def compute_heat_pump_capacity_kw(self, heat_pump_type: HeatingSystem) -> int:
 
         capacity_kw = (
             HEAT_PUMP_CAPACITY_SCALE_FACTOR[heat_pump_type] * self.floor_area_sqm
         )
-        return min(
-            max(capacity_kw, MIN_HEAT_PUMP_CAPACITY_KW[heat_pump_type]),
-            MAX_HEAT_PUMP_CAPACITY_KW[heat_pump_type],
+        return math.ceil(
```
stephanie-chan-cnz, line 385: Also modified in this PR so that household is assigned the higher capacity where in-between bands, and to return an `int` that can be used in a lookup dictionary of heat pump costs (`MEDIAN_COST_GBP_HEAT_PUMP_AIR_SOURCE`, `MEDIAN_COST_GBP_HEAT_PUMP_GROUND_SOURCE`)

simulation/constants.py
```diff
@@ -128,3 +128,25 @@ class EventTrigger(enum.Enum):
...
+    MEDIUM = 1
+    LARGE = 2
+
+
+# Source: https://www.ovoenergy.com/guides/energy-guides/how-much-heating-energy-do-you-use
+# Assume figure of 133kWh/m2a a reflects an average heating system Coefficient of Performance of 0.92 (gas boiler)
+# 133 * 0.92 = 122kWh/m2a
+HEATING_KWH_PER_SQM_ANNUAL = 122
+
+COEFFICIENT_OF_PERFORMANCE: Dict[HeatingSystem, float] = {
```
RyanJenkinson, line 144: I think this perhaps "overloads" the term "Coefficient of Performance" to include the "efficiency" of the boilers as a "coefficient of performance". I really struggled to come up with a name for this term last time, I'm happy to define it as this, or instead to something like `FUEL_KWH_TO_HEAT_KWH`

simulation/costs.py
```diff
@@ -51,3 +64,141 @@
...
+}
+
+HEATING_FUEL_PRICE_GBP_PER_KWH: Dict[HeatingFuel, float] = {
+    HeatingFuel.GAS: 0.038,
+    HeatingFuel.ELECTRICITY: 0.144,
+    HeatingFuel.OIL: 0.032,
+}
+
+
+def get_unit_and_install_costs(household, heating_system):
```
RyanJenkinson, line 151:
```suggestion
def get_unit_and_install_costs(household: "Household", heating_system: HeatingSystem) -> int:
```

RyanJenkinson (changes requested):
Looks good to me! Just some minor changes / questions.

simulation/agents.py
```diff
@@ -222,6 +227,27 @@ def is_heat_pump_suitable(self) -> bool:
             else True
         )
 
+    @property
+    def property_size(self) -> PropertySize:
+
+        # Source: England/Wales EPC
+        FLOOR_AREA_SQM_33RD_PERCENTILE = 66
+        FLOOR_AREA_SQM_66TH_PERCENTILE = 89
```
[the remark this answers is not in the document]
stephanie-chan-cnz: These values include flats - so property size plays a role in sizing boilers in any type of property. I think this fact should be reflected in all flats all having `is_heat_pump_suitable_archetype = False`?

simulation/tests/test_costs.py
```diff
@@ -0,0 +1,130 @@
...
+        "walls_energy_efficiency": 3,
+        "windows_energy_efficiency": 3,
+        "roof_energy_efficiency": 3,
+        "is_heat_pump_suitable_archetype": True,
+        "is_heat_pump_aware": True,
+    }
+    return Household(**{**default_values, **agent_attributes})
+
+
+def model_factory(**model_attributes):
```
RyanJenkinson, line 48: These can be replaced by the `from tests.common import household_factory, model_factory`

RyanJenkinson (approved):

PR #12, opened 2021-12-07
Households can install a heating system

stephanie-chan-cnz:
- Add household method to install a new heating system, and add it to the `household.step()`

### DIFF
--- a/simulation/tests/test_agents.py
+++ b/simulation/tests/test_agents.py
@@ -473,3 +473,22 @@ class TestHousehold:
             household.choose_heating_system(costs, heating_system_hassle_factor=1)
             == HeatingSystem.BOILER_GAS
         )
+
+    @pytest.mark.parametrize("heating_system", list(HeatingSystem))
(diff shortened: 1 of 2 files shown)


simulation/tests/test_agents.py
```diff
@@ -473,3 +473,22 @@ def test_heat_pumps_never_selected_if_hassle_factor_is_one_and_not_currently_ins
...
         )
+
+    @pytest.mark.parametrize("heating_system", list(HeatingSystem))
+    def test_households_install_working_heating_systems_at_model_current_datetime(
+        self,
+        heating_system,
+    ) -> None:
+
+        household = household_factory(
+            heating_system=random.choices(list(HeatingSystem))
```
RyanJenkinson, line 484: Would `random.choices(list(...))` not return a list? Do we not need a `[0]` at the end to ensure it takes a singleton item? I think we kind of get away with it here since we arent checking the initial heating system value at any point in the method or the test.

RyanJenkinson (approved):

PR #13, opened 2021-12-09
Add fuller simulation tracking

stephanie-chan-cnz:
- Add various household collectors for more comprehensive simulation tracking 
- Small modifications to household methods to resolve consistency issues and bugs observed during analysis of simulation output, or to add various household attributes for simulation logging (e.g. prices of heating types seen at a heating decision)
- modify `AgentBasedModel.run()` so agent_callables which evaluate to `False` (e.g. `is_renovating`) are stored as  `False` (and not `None`) in the agent's history

### DIFF
--- a/abm.py
+++ b/abm.py
@@ -84,7 +84,6 @@ class AgentBasedModel:
                 agent_datum = {
                     agent_callable.__name__: agent_callable(agent)
                     for agent_callable in agent_callables
-                    if agent_callable(agent)
                 }
                 agent_data.append(agent_datum)
 
--- a/abm.py
+++ b/abm.py
@@ -84,6 +84,7 @@ class AgentBasedModel:
                 agent_datum = {
                     agent_callable.__name__: agent_callable(agent)
                     for agent_callable in agent_callables
+                    if agent_callable(agent) is not None
                 }
                 agent_data.append(agent_datum)
 
--- a/simulation/agents.py
+++ b/simulation/agents.py
@@ -99,8 +99,12 @@ class Household(Agent):
         self.windows_energy_efficiency = windows_energy_efficiency
         self.is_heat_pump_aware = is_heat_pump_aware
 
-        # Renovation attributes
+        # Household investment decision attributes
         self.is_renovating = False
+        self.renovate_insulation = False
+        self.renovate_heating_system = False
+        self.epc_c_upgrade_costs = {}
+        self.heating_system_total_costs = {}
 
     @property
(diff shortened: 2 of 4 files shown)


abm.py
```diff
@@ -84,7 +84,7 @@ def run(
                 agent_datum = {
                     agent_callable.__name__: agent_callable(agent)
                     for agent_callable in agent_callables
-                    if agent_callable(agent)
+                    if agent_callable(agent) is not None
```
tomwphillips, line 87: Ah, good catch. Can you add a test for this? It's important behaviour.

It should go in [`TestAgentBasedModel`](https://github.com/centrefornetzero/domestic-heating-abm/blob/main/tests/test_abm.py#L17). Perhaps something like `test_run_yields_data_if_it_is_not_none`. You could have two callables, one that returns something that isn't `None` and a second that returns `None`. Then can you can check you only get the first one.

I would make sure the test fails without this change, then check it passes with this change.

tomwphillips (changes requested):
I recommend in the future that you err on the side of more PRs. I'd have broken this up into ~3 (`agent_datum` bug, costs, collectors). Makes for an easier review and cleaner git history.

abm.py
tomwphillips: I tried your test on my machine without your change and it still passed.

I then wrote this test, which still passes without `is not None`:

```python
    def test_run_yields_data_if_it_is_not_none(self) -> None:
        class HouseholdAgent(Agent):
            def step(self, model=None):
                pass

        def callable_returning_none(agent) -> None:
            return None

        def callable_returning_true(agent) -> bool:
            return True

        model = AgentBasedModel()
        model.add_agents([HouseholdAgent(), HouseholdAgent(), HouseholdAgent()])
        num_steps = 10

        history = model.run(
            num_steps=num_steps, agent_callables=[callable_returning_none, callable_returning_true]
        )

        for step_num, step in enumerate(history):
            agents, _ = step
            for agent in agents:
                assert agent == {"callable_returning_true": True}
```

I think this makes sense, because `bool(None) == False`, so it wouldn't be in the dictionary anyway.

If I've understood correctly, you don't need the extra test or `is not None`.

stephanie-chan-cnz: Ah, I've updated the test to test for the population of `False` values. What I was intending to test was that a callable which returns `False` is populated with `{callable_name: False`}` as agent_datum. The previous model behaviour returned as empty dict, b/c `if agent_callable(agent)` evaluates to False in that scenario.

PR #17, opened 2021-12-10
Minor refactoring - renaming to align with ABM naming conventions

stephanie-chan-cnz:
- `AgentBasedModel.steps` -->  `AgentBasedModel.time_steps`
- `AgentBasedModel.UnorderedSpace` -> `AgentBasedModel.GeographicalRegion`
- `CnzAgentBasedModel` --> `DomesticHeatingABM`
- `AgentBasedModel.step()` and `DomesticHeatingABM.step()` --> `.increment_timestep()`
- `Agent.step()` and `Household.step()` --> `.make_decisions()`
- (collectors) `is_first_step()` --> `is_first_timestep()`
- `household.property_value` --> `household.property_value_gbp`

### DIFF
--- a/abm.py
+++ b/abm.py
@@ -56,7 +56,7 @@ class AgentBasedModel:
         for agent in agents:
             self.add_agent(agent)
 
-    def step(self) -> None:
+    def increment_timestep(self) -> None:
         raise NotImplementedError
 
     def run(
@@ -73,7 +73,7 @@ class AgentBasedModel:
 
         for _ in trange(time_steps, desc="Running simulation"):
             try:
-                self.step()
+                self.increment_timestep()
             except NotImplementedError:
                 pass
 
--- a/abm.py
+++ b/abm.py
@@ -80,7 +80,7 @@ class AgentBasedModel:
             agent_data = []
 
             for agent in self.space:
-                agent.step(self)
+                agent.make_decisions(self)
                 agent_datum = {
                     agent_callable.__name__: agent_callable(agent)
                     for agent_callable in agent_callables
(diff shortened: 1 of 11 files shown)


abm.py
```diff
@@ -22,7 +22,7 @@
 History = Iterator[Tuple[List[Dict[str, Any]], Dict[str, Any]]]
 
 
-class UnorderedSpace:
+class GeographicalRegion:
```
tomwphillips, line 25: I don't think this name makes sense, because the agents are all in one giant bag - there's nothing to do with regions.

RyanJenkinson: I am happy with all the other changes but still not sure what to do about this one.
I get that for us it is a `GeographicalRegion`, but it is more general than that. I think it could be named to `DiscreteSpace`, since each item in that space is discrete. I think everything in the `abm.py` should be named quite generally. 

I'm happy to set `GeographicalRegion = DiscreteSpace` in our `simulation/...` though and use it there in the `DomesticHeatingABM` - WDYT?

stephanie-chan-cnz: thanks both - I'll rename it back to `UnorderedSpace` for now.

RyanJenkinson: Sure, for convenience you could set `GeographicalRegions = UnorderedSpace` in `simulation/model.py` just above the `DomesticHeatingABM` and then change [this line](https://github.com/centrefornetzero/domestic-heating-abm/blob/76a372c9828b3a1ba43b4a5a4f1dca7c3d080268/simulation/model.py#L38) to:
```
super().__init__(space=GeographicalRegions)
```
if that helps it make more sense...
But will leave that up to you.

RyanJenkinson (approved):

PR #18, opened 2021-12-10
Relocate various constants used across agents.py to constants.py

stephanie-chan-cnz:


RyanJenkinson (approved):

simulation/constants.py
```diff
@@ -159,3 +159,32 @@ class PropertySize(enum.Enum):
...
+RENO_NUM_INSULATION_ELEMENTS_UPGRADED = {1: 0.76, 2: 0.17, 3: 0.07}
+
+# An amount a house may set aside for work related to home heating and energy efficiency
+# Expressed as a proportion of their total renovation budget (10%)
+HEATING_PROPORTION_OF_RENO_BUDGET = 0.1
+
+# Upper bound on floor area sqm for to be classed as 'Small', by property type / built form
+# As per the segmentation used in Source: BEIS - WHAT DOES IT COST TO RETROFIT HOMES?
+RETROFIT_COSTS_SMALL_PROPERTY_SQM_LIMIT = {
+    "FLAT": 54,
```
RyanJenkinson, line 181: Instead of keys with strings, you could use `InsulationSegment` so it has type `Dict[InsulationSegment, int]`

PR #20, opened 2021-12-13
Add RHI intervention to support simulation validation over historical periods

stephanie-chan-cnz:
In order to reconcile the outputs of the ABM against historical uptake, we need to reflect the conditions of the Domestic RHI in the ABM, which has been live since April 2014 (https://www.ofgem.gov.uk/environmental-and-social-schemes/domestic-renewable-heat-incentive-domestic-rhi). This PR:

- Adds a function to calculate the payments that households would be likely to receive under the RHI
- Passes an intervention label as a command line arg to the model
- Updates a household's method to compute total cost of a given heating system, to subtract any relevant subsidy

### DIFF
--- a/simulation/__main__.py
+++ b/simulation/__main__.py
@@ -52,6 +52,12 @@ def parse_args(args=None):
         help="A value between 0 and 1 which suppresses the likelihood of a household choosing a given heating system (the higher the value, the lower the likelihood)",
     )
 
+    parser.add_argument(
+        "--intervention",
+        choices=["rhi"],
+        type=str,
+    )
+
     return parser.parse_args(args)
 
 
@@ -68,6 +74,7 @@ if __name__ == "__main__":
         args.annual_renovation_rate,
         args.household_num_lookahead_years,
         args.heating_system_hassle_factor,
+        args.intervention,
     )
 
     write_jsonlines(history, args.history_filename)
--- a/simulation/model.py
+++ b/simulation/model.py
@@ -13,6 +13,7 @@ from simulation.constants import (
     ConstructionYearBand,
     Epc,
     HeatingSystem,
+    InterventionType,
     OccupantType,
     PropertyType,
 )
@@ -26,6 +27,7 @@ class DomesticHeatingABM(AgentBasedModel):
         annual_renovation_rate,
         household_num_lookahead_years,
         heating_system_hassle_factor,
+        intervention,
     ):
         self.start_datetime = start_datetime
         self.step_interval = step_interval
@@ -34,6 +36,9 @@ class DomesticHeatingABM(AgentBasedModel):
         self.heating_systems = set(HeatingSystem)
         self.household_num_lookahead_years = household_num_lookahead_years
         self.heating_system_hassle_factor = heating_system_hassle_factor
+        self.intervention = (
+            InterventionType[intervention.upper()] if intervention else None
+        )
 
         super().__init__(UnorderedSpace())
 
@@ -87,6 +92,7 @@ def create_and_run_simulation(
     annual_renovation_rate: float,
     household_num_lookahead_years: int,
     heating_system_hassle_factor: float,
+    intervention: str,
 ):
     model = DomesticHeatingABM(
         start_datetime=start_datetime,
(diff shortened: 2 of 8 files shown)


simulation/model.py
```diff
@@ -34,6 +36,9 @@ def __init__(
         self.heating_systems = set(HeatingSystem)
         self.household_num_lookahead_years = household_num_lookahead_years
         self.heating_system_hassle_factor = heating_system_hassle_factor
+        self.intervention = (
+            InterventionType[intervention.upper()] if intervention else None
+        )
```
RyanJenkinson, line 41: Ok for now, if we only have one intervention type (as we do), but perhaps this becomes a list of potential `InterventionTypes` in the future. Then, in the `Household` agent you could check for `InterventionType.RHI in model.interventions`.

It might be that we change this approach completely for more complicated interventions, and I think its fine for now.

simulation/model.py
```diff
@@ -26,6 +27,7 @@ def __init__(
         annual_renovation_rate,
         household_num_lookahead_years,
         heating_system_hassle_factor,
+        intervention,
```
RyanJenkinson, line 30: It probably makes a bit more sense to pass in a `Optional[InterventionType]` (later perhaps being an `Optional[List[InterventionType])`) here rather than a string (which gets transformed under the hood).

simulation/__main__.py
```diff
@@ -52,6 +52,12 @@ def restrict_between_0_and_1(input_value: float):
         help="A value between 0 and 1 which suppresses the likelihood of a household choosing a given heating system (the higher the value, the lower the likelihood)",
     )
 
+    parser.add_argument(
+        "--intervention",
+        choices=["rhi"],
+        type=str,
```
RyanJenkinson, line 58: What do you think about making this a `type=InterventionType` and using a trick of `choices=list(InterventionType)` to convert it to the enum type right off the bat, rather than converting it inside the agent class?

Reference: https://stackoverflow.com/questions/43968006/support-for-enum-arguments-in-argparse/55500795

simulation/tests/test_costs.py
```diff
@@ -100,3 +101,40 @@ def test_heat_pumps_are_cheaper_to_reinstall_than_install_first_time(
...
+            heating_system=random.choices(list(BOILERS))[0]
+        )
+
+        assert estimate_rhi_annual_payment(household_with_boiler, heat_pump) > 0
+
+    @pytest.mark.parametrize("boiler", set(BOILERS))
+    def test_rhi_annual_payments_zero_for_households_switching_to_boilers(self, boiler):
+
+        household = household_factory(
+            heating_system=random.choices(list(HeatingSystem))[0]
```
RyanJenkinson, line 120: Doesnt the `household_factory` come pre-populated with a `heating_system` so we don't need to sample randomly from one here for this test?

RyanJenkinson (commented):
All the RHI code and tests look great, so I am approving them. Just a quick comment on how we might want to rethink the `argparse` bit. If you want to follow that up in a separate PR then we can do that, and I can approve this PR as-is.

simulation/tests/test_costs.py
stephanie-chan-cnz: It's just so the test gets exposure to more than than the default `heating_system` in `household_factory()`. I've applied this same principle across a lot of tests.

simulation/model.py
stephanie-chan-cnz: Yeah, I think b/c of the slight uncertainty about ever needing to 'layer' interventions, I'm proposing we keep it simple -- just account for the known use case.

simulation/tests/test_costs.py
RyanJenkinson: Ah, ok. If we did want to make the `household_factory()` non-deterministic then we could apply the randomness inside that function, to save you from adding this to a lot of separate tests. Happy either way.

simulation/__main__.py
stephanie-chan-cnz: I'm reluctant to over-engineer the command line args for now, given we might want to fundamentally revisit the way large batches of parameters get passed to a model

RyanJenkinson (approved):
Approved, and will defer `argparse` comments to a later PR.

PR #23, opened 2021-12-16
Add mandatory command line arguments

tomwphillips:
This lays the foundation for a second PR to check random seeds are correctly set.


simulation/__main__.py
```diff
@@ -16,6 +16,9 @@ def convert_to_timedelta(minutes_string):
 
     parser = argparse.ArgumentParser()
 
+    parser.add_argument("household_distribution_file", type=pd.read_csv)
+    parser.add_argument("history_file", default="history.jsonl")
```
stephanie-chan-cnz, line 20: What does it mean for a mandatory arg to have a default value?

stephanie-chan-cnz (dismissed):

simulation/__main__.py
tomwphillips: Good point, this doesn't make sense. I'll remove it.

RyanJenkinson (approved):

PR #26, opened 2021-12-17
Make simulations reproducible

tomwphillips:


simulation/__main__.py
```diff
@@ -65,12 +68,28 @@ def restrict_between_0_and_1(input_value: float):
...
+        datetime.datetime.fromisoformat(string)
+        return string
+
+    parser.add_argument(
+        "--seed",
+        default=datetime.datetime.now().isoformat(),
+        type=is_isoformat_datetime_string,
+        help="""
+        Seed for random number generator. Default is now.
+    """,
```
RyanJenkinson, line 81: This looks like it has formatted incorrectly, will it not fit on one line with `"` ?
```suggestion
        help="Seed for random number generator. Default is now.",
```

simulation/__main__.py
```diff
@@ -65,12 +68,28 @@ def restrict_between_0_and_1(input_value: float):
         help="A factor by which current (2021) air source heat pump unit+install costs will have declined by, as of the end of 2022",
     )
 
+    def is_isoformat_datetime_string(string):
```
RyanJenkinson, line 71: Function name hints at a boolean being returned, but its a string. I believe you are using the `fromisoformat()` method to catch any errors in string formatting. Perhaps the following is easier to read:
```suggestion
    def check_string_is_isoformat_datetime(string) -> str:
```

simulation/__main__.py
```diff
@@ -65,12 +68,28 @@ def restrict_between_0_and_1(input_value: float):
...
+        metavar="YYYY-MM-DD[*HH[:MM[:SS[.fff[fff]]]]",
+    )
+
     return parser.parse_args(args)
 
 
 if __name__ == "__main__":
     args = parse_args()
 
+    random.seed(args.seed)
```
RyanJenkinson, line 91: I think it's good to have a test for `PYTHONHASHSEED='0'` but to enforce it we could have a `setup()` function or something in the `if __name__ == "__main__":` here like:
```
import os
import sys
hashseed = os.getenv('PYTHONHASHSEED')
if not hashseed:
    os.environ['PYTHONHASHSEED'] = '0'
    os.execv(sys.executable, [sys.executable] + sys.argv)
```
source: https://stackoverflow.com/a/64138050

Will leave that thought with you, feel free to resolve if you don't think its useful / too complex.

simulation/model.py
```diff
@@ -68,6 +68,7 @@ def create_household_agents(
 ) -> Iterator[Household]:
     for household in household_population.itertuples():
```
RyanJenkinson, line 69: I think this forces `id` to be a column in the dataframe which isnt strictly necessary. I'd probably prefer something like:
```
for household_id, household in household_population.itertuples(index=True)
``` 
or 
```
for household_id, household in enumerate(household_population.itertuples())
```
and remove the `id` column from the dataframe input requirements.

simulation/__main__.py
RyanJenkinson: If we are going to stick with the current approach (and not do the above, which I am fine with) could we add a `.env.example` with `PYTHONHASHSEED='0'` to the repo and add an instruction to the `README.md` to `cp .env.example .env` before running the code? Thanks.

RyanJenkinson (changes requested):
Generally looks great! Just a few really small things before I approve.

simulation/__main__.py
tomwphillips: No, it's how `black` does it.

simulation/__main__.py
tomwphillips: Yeah, I think it's too complicated.

I'll add it to `.env.template`.

simulation/model.py
tomwphillips: I'm confused by your comment. I _want_ `id` to be a column in the dataframe because I use it. It's simple and straightforward.

RyanJenkinson: My thinking was that when we extract the rows from the data warehouse there wont be an `id` column there (we will have to use the in-built `pandas` index or something else). 
Perhaps we address this in a future PR when we are explicitly dealing with that integration. 
Happy to close this for now and address later, if you want.

tomwphillips: We can add an ID column using `farm_fingerprint`, which will create a 64-bit integer fingerprint of the each record. It'll be a handy way to keep track of the households in the input.

RyanJenkinson: Closing as we can reproducibly generate an ID in BigQuery using `farm_fingerprint`.

RyanJenkinson (approved):

PR #28, opened 2022-01-06
Fix overflow error

RyanJenkinson:
Introduces a cap on the `cost_as_proportion_of_budget` in the `choose_heating_system` method to prevent `math.exp` from overflowing.


tomwphillips:
Let's check with @stephanie-chan-cnz when she's back, but three points:
1) 0 instead of `math.inf`.
2) Reformulate using logs to avoid overflow?
3) Does this matter? I'm not familiar with the surrounding code.

stephanie-chan-cnz:
I suggest we add a simply add a cap for now which prevents overflow, like 50. 

```
multiple_cap = 50
cost_as_proportion_of_budget = min(costs[heating_system] / self.renovation_budget, multiple_cap)
weight = 1 / (1 + cost_as_proportion_of_budget)
```

A value of 50 means a heating system is 50 times out of a household's budget - i.e. an extremely very low weight of choosing it. The choice modelling is rather arbitrary and it is likely we will revise it in some fundamental way after historical reconciliation to RHI numbers, but I think its form should be guided by those outputs (vs spending time further designing it now).

simulation/agents.py
```diff
@@ -444,9 +444,13 @@ def choose_heating_system(
     ):
 
         weights = []
+        multiple_cap = 50  # An arbritrary cap to prevent math.exp overflowing
```
RyanJenkinson, line 447: It seems like the maximum we could use here is `700`. Let me know if you want to change it, or leave it as-is

stephanie-chan-cnz (approved):

PR #32, opened 2022-01-10
Add fuller simulation tracking - prior heating system and functioning status

stephanie-chan-cnz:
- Update of a household's heating system status is moved from `household.install_heating_system()` to `household.update_heating_status()`. This ensures that the model collector for this attribute will obtain value `False` for the step in which a household's heating system has broken down (previously, it is updated back to `True` before the step is completed). A related test around heating system installation is modified accordingly.
- Add household attribute `household_heating_system_prior`, which is initialised to `None`, and populated with a heating system at every time step after a household has changed heating system. This is to facilitate downstream analyses on breakdown events on very large files
- Add collectors for `household_heating_functioning` and `household_heating_system_prior`

### DIFF
--- a/simulation/agents.py
+++ b/simulation/agents.py
@@ -463,7 +463,6 @@ class Household(Agent):
 
         self.heating_system = heating_system
         self.heating_system_install_date = model.current_datetime.date()
-        self.heating_functioning = True
 
     def update_heating_status(self, model: "DomesticHeatingABM") -> None:
 
@@ -479,6 +478,8 @@ class Household(Agent):
         proba_failure = probability_density * step_interval_years
         if random.random() < proba_failure:
             self.heating_functioning = False
+        else:
+            self.heating_functioning = True
 
     def compute_heat_pump_capacity_kw(self, heat_pump_type: HeatingSystem) -> int:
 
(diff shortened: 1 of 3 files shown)


simulation/collectors.py
```diff
@@ -45,6 +45,16 @@ def household_heating_system(household) -> str:
     return household.heating_system.name
 
 
+def household_heating_system_prior(household) -> Optional[str]:
+    return (
+        household.heating_system_prior.name if household.heating_system_prior else None
```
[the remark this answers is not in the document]
stephanie-chan-cnz: yes - as we're extracting `.name` from the enum, which returns an error for `None` cases

simulation/agents.py
```diff
@@ -479,6 +480,8 @@ def update_heating_status(self, model: "DomesticHeatingABM") -> None:
         proba_failure = probability_density * step_interval_years
         if random.random() < proba_failure:
             self.heating_functioning = False
+        else:
+            self.heating_functioning = True
```
RyanJenkinson, line 484: WDYT about replacing this with `self.heating_functioning = not self.true_with_probability(proba_failure)` or `self.true_with_probability(1 - proba_failure)` ?

stephanie-chan-cnz: I feel like the overall step flow (the fact that it gets reset back to functioning) is easier to read in current version

PR #34, opened 2022-01-10
add command line arg to override heat pump suitability to True for all households

stephanie-chan-cnz:
- Setting this arg to `True` means that all households in the simulation will be initialised with `is_heat_pump_suitable_archetype=True`, instead of as per the source file (in which mid-terraced houses and flats are categorised as unsuitable)
- Doing so is important for downstream analyses on the possible ranges of outcomes, as HP suitability is a contested topic -- the source dataset represents a conservative estimate of suitability, while the override represents the most optimistic

### DIFF
--- a/simulation/__main__.py
+++ b/simulation/__main__.py
@@ -62,8 +62,8 @@ def parse_args(args=None):
     )
 
     parser.add_argument(
-        "--override-heat-pump-suitability",
-        default=True,
+        "--all-agents-heat-pump-suitable",
+        default=False,
         type=bool,
         help="When True, 100% of households are assumed suitable for heat pumps. When False, households are assigned a heat pump suitability as per the source data file.",
     )
@@ -108,7 +108,7 @@ if __name__ == "__main__":
         args.heating_system_hassle_factor,
         args.intervention,
         args.air_source_heat_pump_discount_factor_2022,
-        args.override_heat_pump_suitability,
+        args.all_agents_heat_pump_suitable,
     )
 
     write_jsonlines(history, args.history_file)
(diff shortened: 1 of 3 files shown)


simulation/__main__.py
```diff
@@ -61,6 +61,13 @@ def restrict_between_0_and_1(input_value: float):
         type=str,
     )
 
+    parser.add_argument(
+        "--override-heat-pump-suitability",
```
RyanJenkinson, line 65: Can we rename this? I think the `help` string still makes sense with this renaming, but the flag is clearer in terms of its effect on the code.
```suggestion
        "--all-agents-heat-pump-suitable",
```

simulation/__main__.py
```diff
@@ -61,6 +61,13 @@ def restrict_between_0_and_1(input_value: float):
...
+        default=True,
```
RyanJenkinson, line 66: I think `False` is a more sensible default?
```suggestion
        default=False,
```

simulation/model.py
```diff
@@ -92,7 +93,9 @@ def create_household_agents(
             walls_energy_efficiency=household.walls_energy_efficiency,
             windows_energy_efficiency=household.windows_energy_efficiency,
             roof_energy_efficiency=household.roof_energy_efficiency,
-            is_heat_pump_suitable_archetype=household.is_heat_pump_suitable_archetype,
+            is_heat_pump_suitable_archetype=True
+            if override_heat_pump_suitability
```
RyanJenkinson, line 97: Would need to rename this if we change the name

RyanJenkinson (changes requested):

RyanJenkinson (approved):

PR #41, opened 2022-01-11
Update model parameters and heating decision weights post-calibration with RHI uptake

stephanie-chan-cnz:
- This PR modifies a number of default parameters and the weights used when choosing between systems, as part of the process of reconciling the outputs of the ABM with historical heat pump uptake under the RHI. The changes to the latter mean that households weight heating system options with greater discrimination; previously the function weights 'plateaued' when a system was priced at ~40% of a household's budget or lower, while the updated function continues to weight heating systems more highly as they fall below 40% of the budget (i.e. households prefer a system which is 20% of their budget over a heating system which is 30% or 40% of their budget).
- As part of this calibration, it was also found that households who find all heating system options highly unaffordable approximately end up with the same "weight" or uniform likelihood of choosing from the heating systems available. This can lead to the unintuitive outcome that households with very low renovation budgets choose the most expensive options, like heat pumps.
- This therefore also includes a code change to ensure that households for which _all_ options are highly unaffordable (set at x10 out of budget) choose their existing heating system (which can be interpreted as a "repair")

### DIFF
--- a/simulation/agents.py
+++ b/simulation/agents.py
@@ -473,6 +473,11 @@ class Household(Agent):
                 weight *= 1 - heating_system_hassle_factor
             weights.append(weight)
 
+        #  Households for which all options are highly unaffordable (x10 out of budget) "repair" their existing heating system
+        threshold_weight = 1 / (1 + math.exp(10))
+        if all([w < threshold_weight for w in weights]):
+            return self.heating_system
+
         return random.choices(list(costs.keys()), weights)[0]
 
     def install_heating_system(
--- a/simulation/agents.py
+++ b/simulation/agents.py
@@ -468,13 +468,13 @@ class Household(Agent):
             cost_as_proportion_of_budget = min(
                 costs[heating_system] / self.renovation_budget, multiple_cap
             )
-            weight = 1 / (1 + math.exp(cost_as_proportion_of_budget))
+            weight = 1 / math.exp(cost_as_proportion_of_budget)
             if self.is_heating_system_hassle(heating_system):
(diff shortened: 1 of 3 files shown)


simulation/agents.py
```diff
@@ -468,11 +468,16 @@ def choose_heating_system(
             cost_as_proportion_of_budget = min(
                 costs[heating_system] / self.renovation_budget, multiple_cap
             )
-            weight = 1 / (1 + math.exp(cost_as_proportion_of_budget))
+            weight = 1 / math.exp(cost_as_proportion_of_budget)
```
RyanJenkinson, line 471: Could also be this, but I think the current way is easier to read
```suggestion
            weight = math.exp(-cost_as_proportion_of_budget)
```

simulation/agents.py
```diff
@@ -468,11 +468,16 @@ def choose_heating_system(
...
             if self.is_heating_system_hassle(heating_system):
                 weight *= 1 - heating_system_hassle_factor
             weights.append(weight)
 
+        #  Households for which all options are highly unaffordable (x10 out of budget) "repair" their existing heating system
+        threshold_weight = 1 / math.exp(10)
+        if all([w < threshold_weight for w in weights]):
```
RyanJenkinson, line 478: Potential suggestion to make it more readable, but I think what you have is fine, especially with the comment.
```suggestion
        if all([weight < threshold_weight for weight in weights]):
```

RyanJenkinson (approved):
LGTM, just two potential things but non-blocking. Could we update the title of the PR to specifically mention post-calibration _**with RHI uptake**_ ?

PR #42, opened 2022-01-12
Decompose heating system choices

stephanie-chan-cnz:
- This updates the output of `household.get_total_heating_system_costs()` to return a dictionary of unit/install, fuel bill, and subsidy costs (instead of their summed value). In this way, we are better able to decompose simulation outputs downstream -- e.g. understand the primary contributing factors behind the perceived total prices.
- As before, these values are stored as household attributes on steps in which a heating system purchase is made, then reset at `household.update_heating_status()` -- which is the first function run at every household step.
- `household.epc_c_upgrade_costs` is renamed to `household.insulation_element_upgrade_costs` for clarity -- it is a dictionary of upgrade costs by element (roof, windows, walls)
- Associated household collectors are amended/added too

### DIFF
--- a/simulation/agents.py
+++ b/simulation/agents.py
@@ -154,7 +154,7 @@ class Household(Agent):
         self.heating_system_costs_insulation = {}
         self.insulation_element_upgrade_costs = {}
 
-    @propertyheating_system_costs_insulation
+    @property
     def heating_fuel(self) -> HeatingFuel:
         return HEATING_SYSTEM_FUEL[self.heating_system]
 
--- a/simulation/agents.py
+++ b/simulation/agents.py
@@ -147,12 +147,7 @@ class Household(Agent):
         self.renovate_insulation = False
         self.renovate_heating_system = False
         self.boiler_upgrade_grant_available = False
-        self.boiler_upgrade_grant_used = 0
-        self.heating_system_costs_unit_and_install = {}
-        self.heating_system_costs_fuel = {}
-        self.heating_system_costs_subsidies = {}
-        self.heating_system_costs_insulation = {}
-        self.insulation_element_upgrade_costs = {}
+        self.reset_previous_heating_decision_log()
 
     @property
     def heating_fuel(self) -> HeatingFuel:
--- a/simulation/agents.py
+++ b/simulation/agents.py
@@ -458,11 +458,7 @@ class Household(Agent):
         if not model.intervention:
             subsidies = 0
 
-        return {
-            "unit_and_install_costs": unit_and_install_costs,
-            "fuel_costs_net_present_value": fuel_costs_net_present_value,
-            "subsidies": -subsidies,
-        }
+        return unit_and_install_costs, fuel_costs_net_present_value, -subsidies
 
     def choose_heating_system(
         self, costs: Dict[HeatingSystem, float], heating_system_hassle_factor: float
@@ -496,9 +492,9 @@ class Household(Agent):
             if heating_system == HeatingSystem.HEAT_PUMP_GROUND_SOURCE:
                 self.boiler_upgrade_grant_used = 6_000
 
-    def update_heating_status(self, model: "DomesticHeatingABM") -> None:
+    def reset_previous_heating_decision_log(self) -> None:
 
-        # reset any values specific to a previous heating system decision made by household
+        # resets attributes specific to a previous heating system decision
         self.heating_system_costs_unit_and_install = {}
         self.heating_system_costs_fuel = {}
         self.heating_system_costs_subsidies = {}
@@ -506,6 +502,10 @@ class Household(Agent):
         self.insulation_element_upgrade_costs = {}
         self.boiler_upgrade_grant_available = False
         self.boiler_upgrade_grant_used = 0
+
+    def update_heating_status(self, model: "DomesticHeatingABM") -> None:
+
+        self.reset_previous_heating_decision_log()
 
         step_interval_years = model.step_interval / datetime.timedelta(days=365)
         probability_density = weibull_hazard_rate(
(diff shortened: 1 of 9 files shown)


simulation/collectors.py
```diff
@@ -133,43 +133,113 @@ def household_annual_heating_fuel_bill(household) -> int:
...
-    return int(household.heating_system_total_costs.get(HeatingSystem.BOILER_OIL) or 0)
+def household_heating_system_costs_unit_and_install_boiler_electric(household) -> int:
+    return household.heating_system_costs_unit_and_install.get(HeatingSystem.BOILER_GAS)
 
 
-def household_heating_system_total_cost_boiler_electric(household) -> int:
-    return int(
-        household.heating_system_total_costs.get(HeatingSystem.BOILER_ELECTRIC) or 0
+def household_heating_system_costs_unit_and_install_boiler_oil(household) -> int:
+    return household.heating_system_costs_unit_and_install.get(HeatingSystem.BOILER_GAS)
```
RyanJenkinson, line 157: typo?
```suggestion
    return household.heating_system_costs_unit_and_install.get(HeatingSystem.BOILER_OIL)
```

simulation/agents.py
```diff
@@ -548,22 +559,52 @@ def make_decisions(self, model):
...
+            costs_fuel = {}
+            costs_subsidies = {}
+            costs_insulation = {}
+
             for heating_system in heating_system_options:
-                costs[heating_system] = self.get_total_heating_system_costs(
+
+                heating_system_costs = self.get_total_heating_system_costs(
                     heating_system, model
                 )
```
RyanJenkinson, line 571: If we changed the output to a tuple then we wouldnt need to do item assignment below which will take up a lot of time (max 5 times in this for loop, and `make_decisions` is called for each agent step).
```suggestion
                (
                    costs_unit_and_install[heating_system],
                    costs_fuel[heating_system],
                    costs_subsidies[heating_system]
                ) = self.get_total_heating_system_costs(
                    heating_system, model
                )
```

simulation/tests/test_agents.py
```diff
@@ -580,8 +584,12 @@ def test_total_heating_system_costs_are_lower_for_heat_pumps_if_model_interventi
...
         )
 
-        assert household.get_total_heating_system_costs(
-            heat_pump, model_with_boiler_upgrade_scheme
-        ) < household.get_total_heating_system_costs(
-            heat_pump, model_without_boiler_upgrade_scheme
+        assert sum(
+            household.get_total_heating_system_costs(
+                heat_pump, model_with_boiler_upgrade_scheme
+            ).values()
```
RyanJenkinson, line 590: Note that if we changed the output to a tuple we could simply drop `.values()` in all of the tests

simulation/collectors.py
```diff
@@ -225,14 +295,29 @@ def get_agent_collectors(
...
-        household_heating_system_total_cost_heat_pump_air_source,
-        household_heating_system_total_cost_heat_pump_ground_source,
+        household_element_upgrade_cost_roof,
+        household_element_upgrade_cost_walls,
+        household_element_upgrade_cost_windows,
+        household_heating_system_costs_unit_and_install_boiler_gas,
+        household_heating_system_costs_unit_and_install_boiler_electric,
+        household_heating_system_costs_unit_and_install_boiler_oil,
+        household_heating_system_costs_unit_and_install_heat_pump_air_source,
+        household_heating_system_costs_unit_and_install_heat_pump_ground_source,
```
RyanJenkinson, line 305: I remember when we had lots of `agent_collectors` last time it made the simulation quite slow. One way around that this time would be to have a single collector `household_heating_system_costs_unit_and_install` (and similar for the below groups) that extracts the full dictionary.

However, the main "issue" with that idea is then we are working with `pd.DataFrame`'s that have dictionary objects in the values of the dataframe, which is really messy and would require postprocessing.

I'm not sure if there is a better way to handle this, and perhaps we just leave it for now.

RyanJenkinson (changes requested):
I think theres a couple of typos that need fixing, but perhaps I'm wrong.

Also, just added a couple of thoughts in the review, too. Feel free to accept / reject as needed.

simulation/agents.py
```diff
@@ -146,10 +146,13 @@ def __init__(
...
         self.renovate_heating_system = False
-        self.epc_c_upgrade_costs = {}
-        self.heating_system_total_costs = {}
         self.boiler_upgrade_grant_available = False
         self.boiler_upgrade_grant_used = 0
+        self.heating_system_costs_unit_and_install = {}
+        self.heating_system_costs_fuel = {}
+        self.heating_system_costs_subsidies = {}
+        self.heating_system_costs_insulation = {}
+        self.insulation_element_upgrade_costs = {}
```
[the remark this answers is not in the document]
stephanie-chan-cnz: `self.reset_previous_heating_decision_log()`?

simulation/collectors.py
stephanie-chan-cnz: Yeah, indeed - I'd prefer a flat format over dictionaries-in-dataframes, which is why it's this way for now.

simulation/agents.py
```diff
@@ -496,9 +492,9 @@ def install_heating_system(
             if heating_system == HeatingSystem.HEAT_PUMP_GROUND_SOURCE:
                 self.boiler_upgrade_grant_used = 6_000
 
-    def update_heating_status(self, model: "DomesticHeatingABM") -> None:
+    def reset_previous_heating_decision_log(self) -> None:
```
RyanJenkinson, line 500: You could now call this in the `__init__` too if you wanted to? But happy as is.

RyanJenkinson (dismissed):

RyanJenkinson (approved):

PR #43, opened 2022-01-12
Update prices of fuel per kWh used across simulation

stephanie-chan-cnz:
- Revises fuel pence per kwh estimates to align with [Energy Savings Trust ](https://energysavingtrust.org.uk/about-us/our-data/ (England, Scotland and Wales)) estimates, which were last updated in November 2021 and based on predicted fuel prices for 2022

### DIFF
--- a/simulation/costs.py
+++ b/simulation/costs.py
@@ -147,10 +147,12 @@ MEAN_COST_GBP_BOILER_ELECTRIC: Dict[PropertySize, int] = {
     PropertySize.LARGE: 2250 + BOILER_INSTALLATION_COST_GBP,
 }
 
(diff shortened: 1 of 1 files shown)


RyanJenkinson (approved):
Sad to see electricity prices go up so much - but they're in line with what I'm paying at the moment!

RyanJenkinson (commented):
Do we want to add in a boolean flag `--use-high-fuel-prices` that toggles between the higher prices we are currently seeing (in this PR), and our previous "typical" values. 
These new values, which were taken during a "global gas crisis" would otherwise be applied throughout the full simulation (2022 - 2032) and I don't think that is what we want to do?

PR #46, opened 2022-01-13
Add `--bigquery` argument

tomwphillips:


simulation/tests/test_main.py
```diff
@@ -79,6 +80,35 @@ def test_help_flag(self):
...
+        monkeypatch.setattr(simulation.__main__.pd, "read_gbq", mock_read_gbq)
+
+        args = parse_args([output_file, "--bigquery", "select * from table"])
+        pd.testing.assert_frame_equal(args.households, MOCK_QUERY_RESULT)
+        assert args.history_file == output_file
+
+    def test_bigquery_argument_and_households_file_are_mutually_exclusive(
+        self, households_file, output_file
+    ):
+        with pytest.raises(SystemExit):
```
RyanJenkinson, line 98: The test is failing on this test only, because it is exiting with an `OSError`. I think it's because you need to `monkeypatch` the `pd.read_gbq` method as above. I have confirmed this change causes the test to pass for me locally, though there might be another way.
```suggestion
        def mock_read_gbq(query):
            return pd.DataFrame()

        monkeypatch.setattr(simulation.__main__.pd, "read_gbq", mock_read_gbq)
        
        with pytest.raises(SystemExit)
```

RyanJenkinson (changes requested):
PR looks good, just a change to fix the failing test.

RyanJenkinson (approved):

PR #47, opened 2022-01-14
Update cl arg --intervention to pass multiple interventions to simulation

stephanie-chan-cnz:
- Update `--intervention` arg to enable multi-intervention simulations in the future, such as Boiler Upgrade Scheme + Boiler Ban
- As not all interventions we plan to implement will result in a subsidy seen by households (e.g. boiler ban), the logic used in `household.get_total_heating_system_costs()` is amended to anticipate this (previously, a £0 subsidy would only be assigned if model intervention attribute was empty)

### DIFF
--- a/simulation/__main__.py
+++ b/simulation/__main__.py
@@ -71,8 +71,8 @@ def parse_args(args=None):
     parser.add_argument(
         "--intervention",
         action="append",
-        help="Valid interventions are: " + ', '.join(member.name for member in InterventionType)
-'RHI, BOILER_UPGRADE_SCHEME',
+        help="Valid interventions are: "
+        + ", ".join(member.name for member in InterventionType),
         type=map_string_to_intervention_type_enum,
     )
 
--- a/simulation/__main__.py
+++ b/simulation/__main__.py
@@ -21,11 +21,8 @@ def parse_args(args=None):
             return float(value)
         raise ValueError(f"Value must be between 0 and 1, got {value}")
 
-    def convert_str_to_intervention_list(intervention_string):
-        return [
-            InterventionType[intervention.upper()]
-            for intervention in intervention_string.split(",")
-        ]
+    def map_string_to_intervention_type_enum(intervention):
+        return InterventionType[intervention.upper()]
 
     parser = argparse.ArgumentParser()
 
@@ -73,9 +70,10 @@ def parse_args(args=None):
     )
 
     parser.add_argument(
-        "--interventions",
-        help="A comma separated list of interventions, without spaces. Valid list items are: rhi, boiler_upgrade_scheme.",
-        type=convert_str_to_intervention_list,
+        "--intervention",
+        action="append",
+        help="Valid interventions are: rhi, boiler_upgrade_scheme.",
+        type=map_string_to_intervention_type_enum,
     )
 
     parser.add_argument(
@@ -123,7 +121,7 @@ if __name__ == "__main__":
         args.annual_renovation_rate,
         args.household_num_lookahead_years,
         args.heating_system_hassle_factor,
-        args.interventions,
+        args.intervention,
         args.air_source_heat_pump_discount_factor_2022,
         args.all_agents_heat_pump_suitable,
     )
--- a/simulation/__main__.py
+++ b/simulation/__main__.py
@@ -71,7 +71,8 @@ def parse_args(args=None):
     parser.add_argument(
         "--intervention",
         action="append",
-        help="Valid interventions are: rhi, boiler_upgrade_scheme.",
+        help="Valid interventions are: " + ', '.join(member.name for member in InterventionType)
+'RHI, BOILER_UPGRADE_SCHEME',
         type=map_string_to_intervention_type_enum,
     )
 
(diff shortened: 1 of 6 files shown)


tomwphillips (changes requested):
Can you add a test for the `--interventions` arg please?

Also, I would recommend using the keyword argument [`action='append'`](https://docs.python.org/3/library/argparse.html#action) which allows you to specify the same argument multiple times and stores the results in a list.

I think you could combine this with `type=some_func_that_maps_string_to_enum` to get a list of enum values.

simulation/agents.py
```diff
@@ -437,19 +437,19 @@ def get_total_heating_system_costs(
...
-        if model.intervention == InterventionType.BOILER_UPGRADE_SCHEME:
+        if InterventionType.BOILER_UPGRADE_SCHEME in model.interventions:
             subsidies = estimate_boiler_upgrade_scheme_grant(heating_system, model)
             if subsidies > 0:
                 self.boiler_upgrade_grant_available = True
 
-        if model.intervention == InterventionType.RHI:
+        elif InterventionType.RHI in model.interventions:
             rhi_annual_payment = estimate_rhi_annual_payment(self, heating_system)
             subsidies = discount_annual_cash_flow(
```
RyanJenkinson, line 447: I think this is fine for our use case, because all the scenarios we want to run won't have conflicting subsidies where the agents will have to "choose which subsidy applies to them", right?

Technically in the current setup you could run RHI and BUS at the same time, but because they apply over non-intersecting time windows I don't think it matters. 

Also we were only using RHI for calibration. In our actual scenario setup we will run {Boiler Ban, BUS, BUS + Policy Costs, BUS + Policy Costs + Boiler Ban}. I think, handily, none of these are conflicting in terms of competing subsidies: in the latter scenario the first intervention impacts subsidies, the second intervention impacts running costs and the third intervention impacts heating system availability.

So I think this implementation is OK, but just wanted a comment here to check you agree there wouldn't be any weird behaviour?

simulation/model.py
```diff
@@ -123,7 +121,7 @@ def create_and_run_simulation(
     annual_renovation_rate: float,
     household_num_lookahead_years: int,
     heating_system_hassle_factor: float,
-    intervention: str,
+    interventions: List[InterventionType],
```
RyanJenkinson, line 124: I think this is technically true, but don't think it matters too much.
```suggestion
    interventions: Optional[List[InterventionType]],
```

RyanJenkinson (approved):
I think this works well for our use case, just a quick comment to double check you agree there isnt any weird behaviour from the simulations we want to run.

Approving as to not block merging.

simulation/agents.py
stephanie-chan-cnz: Exactly! Upfront subsidies only apply for the RHI and BUS, which happen to be mutually exclusive.

simulation/__main__.py
```diff
@@ -66,15 +70,16 @@ def float_between_0_and_1(value: str):
 
     parser.add_argument(
         "--intervention",
-        choices=["rhi", "boiler_upgrade_scheme"],
-        type=str,
+        action="append",
+        help="Valid interventions are: rhi, boiler_upgrade_scheme.",
```
tomwphillips, line 74: You could do this so we don't have to update it every time we add an intervention.

```suggestion
        help="Valid interventions are: " + ', '.join(member.name for member in InterventionType)
'RHI, BOILER_UPGRADE_SCHEME',
```

simulation/__main__.py
```diff
@@ -84,7 +84,7 @@ def convert_str_to_intervention_list(intervention_string):
         "--all-agents-heat-pump-suitable",
         default=False,
         type=bool,
-        help="When True, 100%% of households are assumed suitable for heat pumps. When False, households are assigned a heat pump suitability as per the source data file.",
+        help="When True, 100% of households are assumed suitable for heat pumps. When False, households are assigned a heat pump suitability as per the source data file.",
```
tomwphillips: This isn't a typo. `argparse` lets you use [format specifiers](https://docs.python.org/3/library/argparse.html#help) so to get a `%` character.

tomwphillips:
Thanks for adding the test.

simulation/__main__.py
stephanie-chan-cnz: Yeah I realised that afterwards. This was since updated to `pc` though, which I think reads better in the codebase.

simulation/model.py
```diff
@@ -37,9 +37,7 @@ def __init__(
         self.heating_systems = set(HeatingSystem)
         self.household_num_lookahead_years = household_num_lookahead_years
         self.heating_system_hassle_factor = heating_system_hassle_factor
-        self.intervention = (
-            InterventionType[intervention.upper()] if intervention else None
-        )
+        self.interventions = interventions or []
```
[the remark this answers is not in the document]
stephanie-chan-cnz: Interventions are optional. The logic to check an intervention is in `model.interventions` will fail if it's None. I'll add typing to `DomesticHeatingABM` to clarify this.

tomwphillips (approved):

PR #50, opened 2022-01-14
Add intervention - gas and oil boiler ban

stephanie-chan-cnz:
- Adds gas and oil boiler ban, as of the `--gas-oil-boiler-ban-date` cl arg (default: 2035-01-01)
- Instead of being a static attribute set at model initialisation, the set of heating systems available is revised into a model property, which gets re-evaluated (based upon the model interventions list, boiler ban date and current model datetime) whenever households make heating decisions
- Note that post-ban, some households can have no heating systems available to them -- e.g. due to unsuitability for heat pumps or electric boilers, and/or because the replacement is triggered by `EventTrigger.BREAKDOWN`, in which heat pumps are not available due to the high lead times required. As for the case where households find all heating system options highly unaffordable, the fallback behaviour here is for a household to select their current heating system, which can be interpreted as repairing their heating system or violating a ban in the absence of a viable alternative. This behaviour can be reviewed further separately of this PR.

### DIFF
--- a/simulation/__main__.py
+++ b/simulation/__main__.py
@@ -108,7 +108,7 @@ def parse_args(args=None):
 
     parser.add_argument(
         "--gas-oil-boiler-ban-date",
-        default=datetime.datetime(2035, 1, 1, 0, 0),
+        default=datetime.datetime(2035, 1, 1),
         type=convert_to_datetime,
     )
 
--- a/simulation/__main__.py
+++ b/simulation/__main__.py
@@ -104,6 +104,12 @@ def parse_args(args=None):
         metavar="YYYY-MM-DD[*HH[:MM[:SS[.fff[fff]]]]",
     )
 
+    parser.add_argument(
+        "--gas-oil-boiler-ban-date",
+        default=datetime.datetime(2035, 1, 1, 0, 0),
+        type=convert_to_datetime,
+    )
+
     return parser.parse_args(args)
 
 
@@ -123,6 +129,7 @@ if __name__ == "__main__":
         args.intervention,
         args.air_source_heat_pump_discount_factor_2022,
         args.all_agents_heat_pump_suitable,
+        args.gas_oil_boiler_ban_date,
     )
 
     write_jsonlines(history, args.history_file)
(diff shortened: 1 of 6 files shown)


simulation/__main__.py
```diff
@@ -104,6 +104,12 @@ def check_string_is_isoformat_datetime(string) -> str:
         metavar="YYYY-MM-DD[*HH[:MM[:SS[.fff[fff]]]]",
     )
 
+    parser.add_argument(
+        "--gas-oil-boiler-ban-date",
+        default=datetime.datetime(2035, 1, 1, 0, 0),
```
tomwphillips, line 109: It's 0 by default.

```suggestion
        default=datetime.datetime(2035, 1, 1),
```

tomwphillips: Another thought: should this be `date` rather than `datetime`?

simulation/tests/test_agents.py
```diff
@@ -590,3 +590,26 @@ def test_total_heating_system_costs_are_lower_for_heat_pumps_if_model_interventi
...
+
+    @pytest.mark.parametrize("event_trigger", set(EventTrigger))
+    def test_gas_and_oil_boilers_are_not_in_heating_options_if_gas_oil_ban_intervention_active(
+        self, event_trigger
+    ):
+
+        household = household_factory(heating_system=random.choices(list(BOILERS))[0])
+
+        model_with_gas_oil_boiler_ban = model_factory(
+            start_datetime=datetime.datetime(2035, 3, 1, 0, 0),
```
tomwphillips, line 602:
```suggestion
            start_datetime=datetime.datetime(2035, 3, 1),
```

simulation/tests/test_agents.py
```diff
@@ -590,3 +590,26 @@ def test_total_heating_system_costs_are_lower_for_heat_pumps_if_model_interventi
...
+            interventions=[InterventionType.GAS_OIL_BOILER_BAN],
+            gas_oil_boiler_ban_datetime=datetime.datetime(2030, 1, 1, 0, 0),
```
tomwphillips, line 604:
```suggestion
            gas_oil_boiler_ban_datetime=datetime.datetime(2030, 1, 1),
```

tomwphillips (commented):
LGTM. Only suggestions about a `datetime.datetime` object - batch 'em up and commit if you want.

tomwphillips (approved):

PR #51, opened 2022-01-14
Specify project ID

tomwphillips:


simulation/__main__.py
```diff
@@ -31,7 +33,7 @@ def map_string_to_intervention_type_enum(intervention):
     households.add_argument(
         "--bigquery",
         help="Generate household agents from BigQuery result.",
-        type=pd.read_gbq,
+        type=partial(pd.read_gbq, project_id=os.getenv("PROJECT_ID")),
```
RyanJenkinson, line 36: I see in your commit comment it works locally because `gcloud auth` uses a different authentication
mechanism.

Does this change here mean we need to add an extra `.env` variable of `PROJECT_ID` in order to use the `--bigquery` flag, or will it still pick this up without it if `project_id=None`?

Basically, do we need to update the `.env.template` to include `PROJECT_ID='project_id' # Specify this to use the --bigquery CLI arg`?

tomwphillips: >Basically, do we need to update the .env.template to include PROJECT_ID='project_id' # Specify this to use the --bigquery CLI arg?

No, you don't.

RyanJenkinson (approved):

PR #56, opened 2022-01-17
Support GCS URIs for history file

tomwphillips:


abm.py
```diff
@@ -113,16 +114,14 @@ def wrapper(*args: Any, **kwargs: Any) -> Optional[T]:
     return collect_when_decorator
 
 
-def write_jsonlines(history: History, filename: str) -> None:
-    with open(filename, "w") as file:
-        for step in history:
-            file.write(json.dumps(step, default=str) + "\n")
+def write_jsonlines(history: History, file: TextIO) -> None:
+    for step in history:
```
RyanJenkinson, line 118: Changing `filename: str` to `file: TextIO` is fine with me (though will need to be changed downstream when analysing results in any notebooks).

Another way we could do this is just replace `open(...)` in each of `write_jsonlines` and `read_jsonlines` with `smart_open.open(...)` so that the input is still a `filename` and the API is the same?
But perhaps you are envisioning when working with it in the container this is a better API choice?

simulation/__main__.py
```diff
@@ -134,4 +138,8 @@ def check_string_is_isoformat_datetime(string) -> str:
         args.gas_oil_boiler_ban_date,
     )
 
-    write_jsonlines(history, args.history_file)
+    if args.history_file.startswith("gs://"):
+        history = list(history)
```
RyanJenkinson, line 142: Is there a test for this usecase?

RyanJenkinson (commented):
Just some quick comments, though happy to approve once you have replied.

abm.py
tomwphillips: Makes for a better API. `{write,read}_jsonlines` write and read JSONlines to a file-like object - they don't actually care what it is. They can then work with in-memory streams, files, `smart_open`, whatever.

simulation/__main__.py
tomwphillips: No, what would you test?

PR #58, opened 2022-01-17
Households aware of heat pumps and consider them in breakdowns if boiler ban active

stephanie-chan-cnz:
- Modifies `household.get_heating_system_options()` such that when a gas/boiler ban is active, all (heat pump suitable) households will consider heat pumps -- regardless of the `EventType`, and regardless of their `is_heat_pump_aware` attribute set at household init.

### DIFF
--- a/simulation/agents.py
+++ b/simulation/agents.py
@@ -140,7 +140,7 @@ class Household(Agent):
         self.walls_energy_efficiency = walls_energy_efficiency
         self.roof_energy_efficiency = roof_energy_efficiency
         self.windows_energy_efficiency = windows_energy_efficiency
-        self.is_heat_pump_aware_init = is_heat_pump_aware
+        self.is_heat_pump_aware = is_heat_pump_aware
 
         # Household investment decision attributes
         self.is_renovating = False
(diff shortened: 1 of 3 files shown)


tomwphillips (changes requested):
I'm not keen on changing `is_heat_pump_aware` to `is_heat_pump_aware_init`. 

I think the logic you've put in the new function `is_heat_pump_aware` is in the wrong place. Instead I think you should put it in `get_heating_system_options`, which gets `model` and can therefore find out if there is a ban or not. That function can then decide whether to use the `is_heat_pump_aware` attribute.

Then you could leave `is_heat_pump_aware` as an attribute set at `__init__`.

tomwphillips:
The failing test is also failing on main so I don't think it's anything to do with your branch. I'm looking into the failure now.

stephanie-chan-cnz:
> I'm not keen on changing `is_heat_pump_aware` to `is_heat_pump_aware_init`.
> 
> I think the logic you've put in the new function `is_heat_pump_aware` is in the wrong place. Instead I think you should put it in `get_heating_system_options`, which gets `model` and can therefore find out if there is a ban or not. That function can then decide whether to use the `is_heat_pump_aware` attribute.
> 
> Then you could leave `is_heat_pump_aware` as an attribute set at `__init__`.

OK, I had first thought about implementing it this way, but decided it preferable to actually update the underlying household attribute the logic is referencing - awareness. I don't mind.

tomwphillips (approved):

PR #59, opened 2022-01-18
Fuel prices per kwh as command line args

stephanie-chan-cnz:
- References to the constant `HEATING_FUEL_PRICE_GBP_PER_KWH` are replaced with model attribute `model.fuel_price_gbp_per_kwh`, which is settable from command line args. Associated tests are refactored or added.
- As `household.annual_heating_fuel_bill` now requires `model` as an arg (to obtain fuel prices), it is dropped from the collector list for simplicity. Its value can in any case be derived in downstream analysis from the collected value `household.annual_kwh_heating_demand` * input fuel price (relevant `household.heating system`).
- Refactors the args of `costs.get_heating_fuel_costs_net_present_value()` to accept `model`, instead of `model.household_num_lookahead_years` for clearer code.


simulation/collectors.py
```diff
@@ -129,10 +129,6 @@ def household_annual_kwh_heating_demand(household) -> int:
     return int(household.annual_kwh_heating_demand)
 
 
-def household_annual_heating_fuel_bill(household) -> int:
-    return household.annual_heating_fuel_bill
```
RyanJenkinson, line 133: I guess we are removing this since now the input depends on the model.

Should we add a `@on_first_step(model_price_gbp_per_kwh_{gas|electricity|oil})` collector to the `model_collectors` to enable us to calculate the fuel bills?

Or should we not save it to the output dataframe and instead calculate it as and when needed?

simulation/model.py
```diff
@@ -30,6 +31,9 @@ def __init__(
         interventions: Optional[List[InterventionType]],
         air_source_heat_pump_discount_factor_2022: float,
         gas_oil_boiler_ban_datetime: datetime.datetime,
+        price_gbp_per_kwh_gas,
+        price_gbp_per_kwh_electricity,
+        price_gbp_per_kwh_oil,
```
RyanJenkinson, line 36: Minor point re: typing.
```suggestion
        price_gbp_per_kwh_gas: float,
        price_gbp_per_kwh_electricity: float,
        price_gbp_per_kwh_oil: float,
```

RyanJenkinson (commented):
Looks good! Only one question really and some minor points.

Happy to approve when we have fixed the flaky test and merge, but not approving just yet until we get the test fixed so we don't merge accidentally.

tomwphillips (approved):

tomwphillips (approved):

PR #61, opened 2022-01-19
Take dataframe conversion out of `read_jsonlines`

tomwphillips:


RyanJenkinson (approved):
LGTM!

If I am understanding correctly, it means that if we do come across the flaky test again it should be clearer if it is a problem with the `json` encoder (or not) since we have now split it out into two functions, as per your commit comment.

This change looks good, and lets keep an eye out to see if the test flakes again in the future.

tomwphillips:
Yes, that's right. I'm going to add a property-based test in another PR.

PR #62, opened 2022-01-19
Write pickle to try and identify cause of flaky test

tomwphillips:
Compare pickle and JSON output in the same test. If they both fail, then we know it's a problem with the ABM class. If only the JSON fails, but the pickle doesn't, then we know it's a problem with the JSON (de)serialization. If the pickle fails but the JSON doesn't, well, I'm not sure what that means. Let's hope that doesn't happen!

### DIFF
--- a/simulation/__main__.py
+++ b/simulation/__main__.py
@@ -1,6 +1,7 @@
 import argparse
 import datetime
 import os
+import pickle
 import random
 import uuid
 from functools import partial
@@ -143,8 +144,10 @@ if __name__ == "__main__":
         args.gas_oil_boiler_ban_date,
     )
 
-    if args.history_file.startswith("gs://"):
-        history = list(history)
+    history = list(history)
+
+    with open(args.history_file + ".pkl", "wb") as file:
+        pickle.dump(history, file)
 
     with smart_open.open(args.history_file, "w") as file:
         write_jsonlines(history, file)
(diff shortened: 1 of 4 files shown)


simulation/__main__.py
```diff
@@ -143,8 +144,10 @@ def check_string_is_isoformat_datetime(string) -> str:
         args.gas_oil_boiler_ban_date,
     )
 
-    if args.history_file.startswith("gs://"):
-        history = list(history)
+    history = list(history)
+
+    with open(args.history_file + ".pkl", "wb") as file:
```
RyanJenkinson, line 149: `args.history_file = output.jsonl` so `args.history_file + ".pkl" = output.jsonl.pkl`. I know this will work, but a confusing naming convention - did you mean:
```suggestion
    with open(args.history_file.replace("jsonl", "pkl") as file
```
or do you want to leave it as-is?

Note: we don't actually check that `args.history_file` ends in `jsonl` anywhere, and it clearly should do in order for the output to save/load properly.

RyanJenkinson (commented):
Code & Idea looks good - just a quick comment before I approve.

simulation/__main__.py
tomwphillips: No, it was intentional. It's only for debugging.

>Note: we don't actually check that args.history_file ends in jsonl anywhere, and it clearly should do in order for the output to save/load properly.

It doesn't actually - the extension doesn't matter.

RyanJenkinson (approved):

PR #63, opened 2022-01-19
Use dict instead of set in UnorderedSpace

tomwphillips:


RyanJenkinson (approved):
Fingers crossed this fixes the test!

PR #64, opened 2022-01-19
Air source heat pump costs change according to specified date schedule

stephanie-chan-cnz:
- Deprecates `--air-source-heat-pump-discount-factor-2022` and passes `--air-source-heat-pump-price-discount-date` as command line arg instead, which allows for the user to specify any number of `YYYY-MM-DD:price_discount` inputs, to reflect some forecast of air source heat pump price changes
- Adds `model.end_datetime` arg, which is derived from `start_date` and `step_interval` model attributes and the `num_steps` arg passed at runtime. This is used in the below method.
- Adds model method `model.get_heat_pump_price_discount_schedule`, which is a dictionary of {YYYY-MM-DD:price_discount} inclusive of the simulation start and end dates. When households make heating decisions, the model property `model.air_source_heat_pump_discount_factor` infers the discount applicable to the current datetime, linearly interpolating across a discount factor schedule where applicable.

### DIFF
--- a/simulation/__main__.py
+++ b/simulation/__main__.py
@@ -129,7 +129,8 @@ def parse_args(args=None):
         "--air-source-heat-pump-price-discount-date",
         action="append",
         type=map_string_to_datetime_float_tuple,
-        help="A factor by which heat pump prices will fall by a specified date, provided in format 'YYYY-MM-DD:price_discount' (e.g. '2023-01-01:0.3')",
+        help="A factor by which heat pump prices will fall by a specified date.",
+        metavar="YYYY-MM-DD:price_discount",
     )
 
     return parser.parse_args(args)
--- a/simulation/__main__.py
+++ b/simulation/__main__.py
@@ -28,6 +28,10 @@ def parse_args(args=None):
     def map_string_to_intervention_type_enum(intervention):
         return InterventionType[intervention.upper()]
 
+    def map_string_to_datetime_float_tuple(date_price_discount_string):
+        date, price_discount = date_price_discount_string.split(":")
+        return datetime.datetime.strptime(date, "%Y-%m-%d"), float(price_discount)
+
     parser = argparse.ArgumentParser()
 
     households = parser.add_mutually_exclusive_group(required=True)
@@ -121,6 +125,13 @@ def parse_args(args=None):
         type=convert_to_datetime,
     )
 
+    parser.add_argument(
+        "--heat-pump-price-discount-date",
+        action="append",
+        type=map_string_to_datetime_float_tuple,
+        help="A factor by which heat pump prices will fall by a specified date, provided in format 'YYYY-MM-DD:price_discount' (e.g. '2023-01-01:0.3')",
+    )
+
     return parser.parse_args(args)
 
 
@@ -141,6 +152,7 @@ if __name__ == "__main__":
         args.air_source_heat_pump_discount_factor_2022,
         args.all_agents_heat_pump_suitable,
         args.gas_oil_boiler_ban_date,
+        args.heat_pump_price_discount_date,
     )
 
     if args.history_file.startswith("gs://"):
--- a/simulation/__main__.py
+++ b/simulation/__main__.py
@@ -98,13 +98,6 @@ def parse_args(args=None):
         help="When True, 100pc of households are assumed suitable for heat pumps. When False, households are assigned a heat pump suitability as per the source data file.",
     )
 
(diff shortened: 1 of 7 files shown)


simulation/__main__.py
```diff
@@ -121,6 +118,13 @@ def check_string_is_isoformat_datetime(string) -> str:
         type=convert_to_datetime,
     )
 
+    parser.add_argument(
+        "--air-source-heat-pump-price-discount-date",
+        action="append",
+        type=map_string_to_datetime_float_tuple,
+        help="A factor by which heat pump prices will fall by a specified date, provided in format 'YYYY-MM-DD:price_discount' (e.g. '2023-01-01:0.3')",
```
tomwphillips, line 125: Use [`metavar`](https://docs.python.org/3/library/argparse.html#metavar) for things like this.

```suggestion
        help="A factor by which heat pump prices will fall by a specified date (e.g. '2023-01-01:0.3')",
        metavar="YYYY-MM-DD:price_discount"
```

simulation/model.py
```diff
@@ -24,6 +24,7 @@ def __init__(
         self,
         start_datetime: datetime.datetime,
         step_interval: int,
+        end_datetime: datetime.datetime,
```
tomwphillips, line 28: I'm not sure about this. Right now in `main` what determines the end date of the simulation is `start_date + (num_steps * step_interval)`, where `start_date` and `step_interval` are properties of `DomesticHeatingABM` and `num_steps` is an argument to `DomesticHeatingABM.run`.

I think it's confusing to have that and `end_datetime`, because `end_datetime` suggests the simulation would stop on that date but in fact it's dependent on `num_steps` and `step_interval`. What happens if `run` runs the simulation past `end_datetime`?

simulation/model.py
```diff
@@ -84,6 +90,34 @@ def increment_timestep(self):
...
+            List[Tuple[datetime.datetime, float]]
+        ],
+    ):
+
+        if heat_pump_price_discount_schedule:
+
+            discount_schedule = {
+                schedule[0]: schedule[1]
+                for schedule in heat_pump_price_discount_schedule
+            }
```
tomwphillips, line 108: A small suggestion - it's more readable to unpack the tuple instead of using indexing, e.g.

```
discount_schedule = {date: discount for date, discount in heat_pump_price_discount_schedule}
```

simulation/model.py
```diff
@@ -84,6 +90,34 @@ def increment_timestep(self):
...
+
+            price_change_dates = discount_schedule.keys()
+            first_date, last_date = min(price_change_dates), max(price_change_dates)
```
tomwphillips, line 111: You don't need to call `.keys()` on the dictionary - `max` and `min` operate on the keys anyway, for example:

```
>>> x = {'a': 3, 'b': 2, 'c': 1}
>>> max(x)
'c'
>>> min(x)
'a'
```

So you can just use `min(discount_schedule), max(discount_schedule)`.

simulation/model.py
```diff
@@ -59,13 +64,19 @@ def heating_systems(self) -> Set[HeatingSystem]:
...
+
+        for idx, date_range in enumerate(date_ranges):
+            if date_range[idx] <= self.current_datetime <= date_range[idx + 1]:
+                proportion_through_range = (self.current_datetime - date_range[0]) / (
+                    date_range[1] - date_range[0]
+                )
+                return (
+                    factors[idx]
+                    + (factors[idx + 1] - factors[idx]) * proportion_through_range
+                )
```
tomwphillips: Just a note that I'm not keen on this but need to have a think...

tomwphillips (changes requested):
See comments below - I'm not too sure about these changes. Can we have a chat about it tomorrow?

simulation/model.py
stephanie-chan-cnz: Nothing happens if the simulation runs past this date. It's just used in the derivation of `model.get_heat_pump_price_discount_schedule`, which must contain a key on or after the maximum `model.current_datetime` in its current setup. It isn't passed as a simulation parameter, so can't be misunderstood to affect the simulation in this way?

simulation/model.py
RyanJenkinson: I agree we should use `bisect`. I think its up to us to decide if we want to linearly interpolate the costs between the dates or just have hard filters where it becomes active at a certain date onwards. I can see the pros and cons of both modelling choices.

PR #68, opened 2022-01-20
Time step in months, one by default

tomwphillips:


simulation/__main__.py
```diff
@@ -60,8 +61,9 @@ def format_uuid(str):
 
     parser.add_argument(
         "--step-interval",
-        type=convert_to_timedelta,
+        type=convert_to_months_relativedelta,
         default=datetime.timedelta(minutes=1440),
```
RyanJenkinson, line 65: Should this be:
```suggestion
        default=relativedelta(months=1),
```

RyanJenkinson (changes requested):
LGTM, just a missing default value I think.

RyanJenkinson (approved):

stephanie-chan-cnz:
Doesn't this return `TypeError: unsupported operand type(s) for /: 'relativedelta' and 'datetime.timedelta'` when running the simulation, due to the attempt to divide one by the other in `model.update_heating_status()`?

stephanie-chan-cnz (changes requested):
This change throws `TypeError: unsupported operand type(s) for /: 'relativedelta' and 'datetime.timedelta' when running the simulation, due to the attempt to divide one by the other in `model.update_heating_status()`

simulation/agents.py
```diff
@@ -513,7 +513,9 @@ def update_heating_status(self, model: "DomesticHeatingABM") -> None:
 
         self.reset_previous_heating_decision_log()
 
-        step_interval_years = model.step_interval / datetime.timedelta(days=365)
+        step_interval_years = (
+            model.step_interval.months + (12 * model.step_interval.years) / 12
```
stephanie-chan-cnz, line 517: Bracketing needs to be updated here, order of operations is problematic

stephanie-chan-cnz (changes requested):

simulation/agents.py
tomwphillips: Whoops! 🙈

stephanie-chan-cnz (approved):

PR #69, opened 2022-01-20
Air source heat pump costs change according to specified date schedule

stephanie-chan-cnz:
- Deprecates `model.air_source_heat_pump_discount_factor_2022` and replaces with `model.air_source_heat_pump_price_discount_date`, enabling the user to more flexibly pass in a (stepped) discount schedule for ASHPs

### DIFF
--- a/simulation/__main__.py
+++ b/simulation/__main__.py
@@ -104,7 +104,6 @@ def parse_args(args=None):
         "--air-source-heat-pump-price-discount-date",
         action="append",
         type=map_string_to_datetime_float_tuple,
-        default=[],
         help="A factor by which heat pump prices will fall by a specified date.",
         metavar="YYYY-MM-DD:price_discount",
     )
--- a/simulation/__main__.py
+++ b/simulation/__main__.py
@@ -28,6 +28,10 @@ def parse_args(args=None):
     def map_string_to_intervention_type_enum(intervention):
         return InterventionType[intervention.upper()]
 
+    def map_string_to_datetime_float_tuple(date_price_discount_string):
+        date, price_discount = date_price_discount_string.split(":")
+        return datetime.datetime.strptime(date, "%Y-%m-%d"), float(price_discount)
+
     parser = argparse.ArgumentParser()
 
     households = parser.add_mutually_exclusive_group(required=True)
@@ -97,10 +101,12 @@ def parse_args(args=None):
     )
 
     parser.add_argument(
-        "--air-source-heat-pump-discount-factor-2022",
-        type=float,
-        default=0.1,
-        help="A factor by which current (2021) air source heat pump unit+install costs will have declined by, as of the end of 2022",
+        "--air-source-heat-pump-price-discount-date",
+        action="append",
+        type=map_string_to_datetime_float_tuple,
+        default=[],
+        help="A factor by which heat pump prices will fall by a specified date.",
+        metavar="YYYY-MM-DD:price_discount",
     )
 
     def check_string_is_isoformat_datetime(string) -> str:
@@ -146,12 +152,12 @@ if __name__ == "__main__":
         args.household_num_lookahead_years,
         args.heating_system_hassle_factor,
         args.intervention,
-        args.air_source_heat_pump_discount_factor_2022,
         args.all_agents_heat_pump_suitable,
         args.gas_oil_boiler_ban_date,
         args.price_gbp_per_kwh_gas,
         args.price_gbp_per_kwh_electricity,
         args.price_gbp_per_kwh_oil,
+        args.air_source_heat_pump_price_discount_date,
     )
 
     if args.history_file.startswith("gs://"):
--- a/simulation/model.py
+++ b/simulation/model.py
@@ -1,6 +1,7 @@
 import datetime
 import random
-from typing import Iterator, List, Optional, Set
+from bisect import bisect
+from typing import Iterator, List, Optional, Set, Tuple
 
 import pandas as pd
 
@@ -29,11 +30,13 @@ class DomesticHeatingABM(AgentBasedModel):
         household_num_lookahead_years: int,
         heating_system_hassle_factor: float,
         interventions: Optional[List[InterventionType]],
-        air_source_heat_pump_discount_factor_2022: float,
         gas_oil_boiler_ban_datetime: datetime.datetime,
         price_gbp_per_kwh_gas: float,
         price_gbp_per_kwh_electricity: float,
         price_gbp_per_kwh_oil: float,
+        air_source_heat_pump_price_discount_schedule: List[
+            Tuple[datetime.datetime, float]
+        ],
     ):
         self.start_datetime = start_datetime
         self.step_interval = step_interval
@@ -42,9 +45,6 @@ class DomesticHeatingABM(AgentBasedModel):
         self.household_num_lookahead_years = household_num_lookahead_years
         self.heating_system_hassle_factor = heating_system_hassle_factor
         self.interventions = interventions or []
-        self.air_source_heat_pump_discount_factor_2022 = (
-            air_source_heat_pump_discount_factor_2022
-        )
         self.boiler_upgrade_scheme_cumulative_spend_gbp = 0
         self.gas_oil_boiler_ban_datetime = gas_oil_boiler_ban_datetime
         self.fuel_price_gbp_per_kwh = {
@@ -52,6 +52,9 @@ class DomesticHeatingABM(AgentBasedModel):
             HeatingFuel.ELECTRICITY: price_gbp_per_kwh_electricity,
             HeatingFuel.OIL: price_gbp_per_kwh_oil,
         }
+        self.air_source_heat_pump_price_discount_schedule = sorted(
+            air_source_heat_pump_price_discount_schedule
+        )
 
         super().__init__(UnorderedSpace())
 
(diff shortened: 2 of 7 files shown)


simulation/model.py
```diff
@@ -68,13 +71,20 @@ def heating_systems(self) -> Set[HeatingSystem]:
...
-        if self.current_datetime.year > 2022:
-            return 1 - self.air_source_heat_pump_discount_factor_2022
-        else:
-            month = self.current_datetime.month
-            return 1 - (month / 12 * self.air_source_heat_pump_discount_factor_2022)
+        if self.air_source_heat_pump_price_discount_schedule:
+
+            step_dates, discount_factors = [
+                step for step in zip(*self.air_source_heat_pump_price_discount_schedule)
+            ]
```
RyanJenkinson, line 78: This might be too long to fit on one line, and happy to leave as is. Or, you could use a walrus in the above if statement e.g `if schedule := self.air_source_heat_pump_price_discount_schedule` and then refer to it as `schedule` below (but then our code will only be `PYTHON_VERSION>=3.8`
```suggestion
            step_dates, discount_factors = zip(*self.air_source_heat_pump_price_discount_schedule)
```

simulation/model.py
```diff
@@ -29,11 +30,13 @@ def __init__(
...
         heating_system_hassle_factor: float,
         interventions: Optional[List[InterventionType]],
-        air_source_heat_pump_discount_factor_2022: float,
         gas_oil_boiler_ban_datetime: datetime.datetime,
         price_gbp_per_kwh_gas: float,
         price_gbp_per_kwh_electricity: float,
         price_gbp_per_kwh_oil: float,
+        air_source_heat_pump_price_discount_schedule: List[
+            Tuple[datetime.datetime, float]
+        ],
```
RyanJenkinson, line 39: Can we change this to an `Optional[List[...]]` to avoid passing in `[]` to instantiate it?
```suggestion
        air_source_heat_pump_price_discount_schedule: Optional[List[
            Tuple[datetime.datetime, float]
        ]],
```

simulation/model.py
```diff
@@ -42,16 +45,16 @@ def __init__(
...
         self.boiler_upgrade_scheme_cumulative_spend_gbp = 0
         self.gas_oil_boiler_ban_datetime = gas_oil_boiler_ban_datetime
         self.fuel_price_gbp_per_kwh = {
             HeatingFuel.GAS: price_gbp_per_kwh_gas,
             HeatingFuel.ELECTRICITY: price_gbp_per_kwh_electricity,
             HeatingFuel.OIL: price_gbp_per_kwh_oil,
         }
+        self.air_source_heat_pump_price_discount_schedule = sorted(
+            air_source_heat_pump_price_discount_schedule
+        )
```
RyanJenkinson, line 57: If we have an optional input, then we can change this to:
```suggestion
        ) if air_source_heat_pump_discount_schedule else None
```
I dont even think we need to instantiate an empty list here after the `else` because you check `if self.air_source_heat_pump_discount_schedule` in the `air_source_heat_pump_discount_factor` property.

simulation/model.py
```diff
@@ -142,12 +152,12 @@ def create_and_run_simulation(
     household_num_lookahead_years: int,
     heating_system_hassle_factor: float,
     interventions: Optional[List[InterventionType]],
-    air_source_heat_pump_discount_factor_2022: float,
     all_agents_heat_pump_suitable: bool,
     gas_oil_boiler_ban_datetime: datetime.datetime,
     price_gbp_per_kwh_gas: float,
     price_gbp_per_kwh_electricity: float,
     price_gbp_per_kwh_oil: float,
+    air_source_heat_pump_price_discount_schedule: List[Tuple[datetime.datetime, float]],
```
RyanJenkinson, line 160:
```suggestion
    air_source_heat_pump_price_discount_schedule: Optional[List[Tuple[datetime.datetime, float]]],
```

simulation/tests/common.py
```diff
@@ -45,10 +45,10 @@ def model_factory(**model_attributes):
         "household_num_lookahead_years": 3,
         "heating_system_hassle_factor": 0.7,
         "interventions": [],
-        "air_source_heat_pump_discount_factor_2022": 0,
         "gas_oil_boiler_ban_datetime": datetime.datetime(2035, 1, 1),
         "price_gbp_per_kwh_gas": 0.0465,
         "price_gbp_per_kwh_electricity": 0.2006,
         "price_gbp_per_kwh_oil": 0.0482,
+        "air_source_heat_pump_price_discount_schedule": [],
```
RyanJenkinson, line 52:
```suggestion
        "air_source_heat_pump_price_discount_schedule": None,
```

simulation/tests/test_costs.py
```diff
@@ -155,28 +154,26 @@ def test_rhi_annual_payments_reach_cap_for_large_households(self, heat_pump):
...
+    def test_air_source_heat_pumps_get_cheaper_across_discount_schedule(self):
 
         household = household_factory(heating_system=HeatingSystem.HEAT_PUMP_AIR_SOURCE)
         model = model_factory(
-            start_datetime=datetime.datetime(2022, 1, 1, 0, 0),
-            air_source_heat_pump_discount_factor_2022=0.3,
+            start_datetime=datetime.datetime(2022, 1, 1),
+            step_interval=datetime.timedelta(minutes=1440),
+            air_source_heat_pump_price_discount_schedule=[
+                (datetime.datetime(2022, 1, 2), 0.3),
```
RyanJenkinson, line 164: We could extract out the variable `discount_factor=0.3` here then test below:
```
assert later_quote = (1 - discount_factor) * first_quote
```

simulation/tests/test_model.py
```diff
@@ -54,6 +54,32 @@ def test_increment_timestep_updates_boiler_upgrade_scheme_cumulative_spend_gbp(
         model.increment_timestep()
         assert model.boiler_upgrade_scheme_cumulative_spend_gbp == 22_000
 
+    def test_air_source_heat_pump_discount_factor_is_zero_if_no_discount_schedule_passed(
+        self,
+    ):
+
+        model = model_factory(air_source_heat_pump_price_discount_schedule=[])
```
RyanJenkinson, line 61:
```suggestion
        model = model_factory(air_source_heat_pump_price_discount_schedule=None)
```

simulation/tests/test_model.py
```diff
@@ -54,6 +54,32 @@ def test_increment_timestep_updates_boiler_upgrade_scheme_cumulative_spend_gbp(
...
+        self,
+    ):
+
+        model = model_factory(
+            start_datetime=datetime.datetime(2022, 2, 1),
+            air_source_heat_pump_price_discount_schedule=[
+                (datetime.datetime(2022, 2, 1), 0),
+                (datetime.datetime(2022, 2, 2), 0.3),
+            ],
+            step_interval=datetime.timedelta(minutes=1440),
```
RyanJenkinson, line 75: n.b whenever the other PR is merged we can change all these `step_interval`'s to `dateutil.relativetimedelta(days=1)`

simulation/tests/test_model.py
```diff
@@ -54,6 +54,32 @@ def test_increment_timestep_updates_boiler_upgrade_scheme_cumulative_spend_gbp(
...
+        assert model.air_source_heat_pump_discount_factor == 0
+
+    def test_air_source_heat_pump_discount_factor_changes_when_crosses_discount_schedule_date(
+        self,
+    ):
+
+        model = model_factory(
+            start_datetime=datetime.datetime(2022, 2, 1),
+            air_source_heat_pump_price_discount_schedule=[
+                (datetime.datetime(2022, 2, 1), 0),
```
RyanJenkinson, line 72: Just because we have already tested for 0 in a previous test so we want to make sure its working correctly
```suggestion
                (datetime.datetime(2022, 2, 1), 0.1),
```

simulation/tests/test_model.py
```diff
@@ -54,6 +54,32 @@ def test_increment_timestep_updates_boiler_upgrade_scheme_cumulative_spend_gbp(
...
+                (datetime.datetime(2022, 2, 2), 0.3),
+            ],
+            step_interval=datetime.timedelta(minutes=1440),
+        )
+
+        assert model.air_source_heat_pump_discount_factor == 0
+
+        model.increment_timestep()
+        assert model.air_source_heat_pump_discount_factor == 0.3
```
RyanJenkinson, line 81: This test will now also check it "stays at the last discount factor" even when we go to the next timestep
```suggestion
        assert model.air_source_heat_pump_discount_factor == 0.1

        model.increment_timestep()
        assert model.air_source_heat_pump_discount_factor == 0.3
        
        model.increment_timestep()
        assert model.air_source_heat_pump_discount_factor == 0.3
```

RyanJenkinson (changes requested):
The logic and structure LGTM! Think we should just tweak some of the tests slightly, and the input argument to `None` when we don't have a schedule, rather than an empty list.

I think its an open question about what a sensible default is, but for now I'm swaying towards `None` as the default.

simulation/__main__.py
```diff
@@ -97,10 +101,12 @@ def format_uuid(str):
...
 
     parser.add_argument(
-        "--air-source-heat-pump-discount-factor-2022",
-        type=float,
-        default=0.1,
-        help="A factor by which current (2021) air source heat pump unit+install costs will have declined by, as of the end of 2022",
+        "--air-source-heat-pump-price-discount-date",
+        action="append",
+        type=map_string_to_datetime_float_tuple,
+        default=[],
```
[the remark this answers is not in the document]
stephanie-chan-cnz: It would be a default of `2023-01-01:0.3` (the float is the amount by which the price is discounted, not the proportion of the original price). But the args get appended to this default, so that would be strange in the case we want to override it.

RyanJenkinson (approved):
LGTM!

PR #77, opened 2022-01-24
Households with heat pumps are always able to pick heat pumps

stephanie-chan-cnz:
- HP ownership (sourced from data warehouse) is disconnected from HP awareness (randomised at simulation). This fixes a logical inconsistency whereby households can own heat pumps but not be aware of them. 
- Note that this will have v. low impact on simulations ran to date due to the simulation time frame (< heating system lifetime), and low initial HP ownership.

### DIFF
--- a/k8s/job.jsonnet
+++ b/k8s/job.jsonnet
@@ -35,17 +35,17 @@ local bigquery_arg = [
 ];
 
 [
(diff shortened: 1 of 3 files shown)


tomwphillips (dismissed):
Approved with a small suggestion.

PR #90, opened 2022-01-26
Add structlog messages

tomwphillips:


abm.py
```diff
@@ -90,6 +96,12 @@ def run(
                 for model_callable in model_callables
             }
 
+            logger.info(
+                "step completed",
+                step=step,
+                elapsed_time_seconds=time.time() - start_time,
```
RyanJenkinson, line 102: Thoughts on adding an ETA to the logs? Could also convert to a `datetime.time` format.
```suggestion
                elapsed_time_seconds=time.time() - start_time,
                estimated_time_remaining_seconds=(time_steps - step) * (time.time() - start_time)
```

If we don't do the above, can we add
```
    total_time_steps=time_steps
```
so we can at least see we are on step X / total.

tomwphillips: My thoughts are that it's not essential - someone can do it in another PR if it becomes a requirement.

RyanJenkinson (approved):

PR #95, opened 2022-01-27
Scale Boiler Upgrade Scheme budget for number of households in simulation

stephanie-chan-cnz:


simulation/model.py
```diff
@@ -57,6 +58,7 @@ def __init__(
             if air_source_heat_pump_price_discount_schedule
             else None
         )
+        self.household_count = household_count
```
RyanJenkinson, line 61: Instead of putting this in the `__init__` should we make it a `@property` ? Isn't it just:

```
    @property
    def household_count(self):
        return len(self.space.agents)
```
Don't know how this will affect the test below, though.

simulation/tests/test_costs.py
```diff
@@ -210,10 +215,20 @@ def test_boiler_upgrade_scheme_grant_is_zero_when_grant_cap_exceeded(
         model = model_factory(
             start_datetime=datetime.datetime(2023, 1, 1, 0, 0),
         )
-        model.boiler_upgrade_scheme_cumulative_spend_gbp = 3_000_000
+
```
RyanJenkinson, line 218: I think to make this work with the `@property` you would need to add the following line here:
```
    model.add_agents([Household()] * num_households)
```
I think you used `24_600` in the factory, so you could define `num_households=24_600` in the test here and do it this way - WDYT?

RyanJenkinson (commented):
Just a comment on how we could simplify the code - let me know what you think!

simulation/model.py
stephanie-chan-cnz: There's no need to re-derive this throughout the simulation though? It's static

simulation/tests/test_costs.py
stephanie-chan-cnz: I think I'd need to update a number of tests such that the model has a non-empty set of household agents :/

RyanJenkinson (approved):

simulation/tests/test_costs.py
```diff
@@ -210,10 +217,23 @@ def test_boiler_upgrade_scheme_grant_is_zero_when_grant_cap_exceeded(
...
+        num_households = random.randint(0, 5)
```
RyanJenkinson, line 221: Think we need to change this to prevent a division by zero below.
```suggestion
        num_households = random.randint(1, 5)
```

RyanJenkinson (approved):

PR #96, opened 2022-01-27
Tweak job config

tomwphillips:
- Increase CPU
- Reduce parallism
- Reduce completions


tomwphillips:
logging_user_domestic-heating-abm_step-elapsed-time  95TH PERCENTILE

Quite a substantial improvement in time per step.

PR #100, opened 2022-01-31
Add heat pump installer capacity to cap heat pump uptake

stephanie-chan-cnz:
- Enables user to pass in a factor for annual heat pump installer growth, which in turn with defaults specified for the time taken to install a heat pump, limits the number of heat pumps which can be installed in a given step. In this way, we prevent heat pump uptake  exceeding some plausible/forecasted heat pump installer capacity.
- With the defaults provided, the implied heat pump installation capacity is 3,400 installs/month or 40k installs/year, which approximately aligns with estimates of 35k currently being installed in the UK per year ([page 8](https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/1026356/domestic-offgg-consultation.pdf))
- Similarly to the Boiler Upgrade Scheme budget, the heat pump installer count is scaled to the actual simulated population to avoid non-meaningful scenarios (e.g. 1,000 agents having access to 1,700 heat pump installers).

### DIFF
--- a/simulation/__main__.py
+++ b/simulation/__main__.py
@@ -121,6 +121,13 @@ def parse_args(args=None):
         metavar="YYYY-MM-DD:price_discount",
     )
 
+    parser.add_argument(
+        "--heat-pump-installer-annual-growth-rate",
+        type=float,
+        default=0,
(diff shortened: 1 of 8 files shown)


tomwphillips (approved):
LGTM.

PR #101, opened 2022-02-01
Add floor for ASHP cost at price of a gas boiler

stephanie-chan-cnz:
- Floors the price of an air source heat pump to the price that a given household would pay for a gas boiler. Note that all decommissioning costs sit independently of projected unit costs - therefore for a household _switching_ heating system, they will still experience these decommissioning costs on top of the floored price.
- Also refactors the function `costs.get_unit_and_install_costs()` for clearer code in the process.

### DIFF
--- a/simulation/tests/test_costs.py
+++ b/simulation/tests/test_costs.py
@@ -10,6 +10,7 @@ from simulation.constants import (
     HeatingSystem,
 )
 from simulation.costs import (
+    MEAN_COST_GBP_BOILER_GAS,
     estimate_boiler_upgrade_scheme_grant,
     estimate_rhi_annual_payment,
     get_heating_fuel_costs_net_present_value,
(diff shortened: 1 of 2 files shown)


simulation/tests/test_costs.py
```diff
@@ -257,3 +258,42 @@ def test_boiler_upgrade_scheme_grant_is_non_zero_for_heat_pumps_when_grant_is_ac
...
+            ],
+        )
+        household = household_factory(heating_system=HeatingSystem.BOILER_GAS)
+        heat_pump_cost = get_unit_and_install_costs(
+            household, HeatingSystem.HEAT_PUMP_AIR_SOURCE, model
+        )
+        heat_pump_price_floor = MEAN_COST_GBP_BOILER_GAS[household.property_size]
+
+        # Decommissioning costs for changing heating system type are randomly generated between 500-2_000 GBP
+        decommission_cost_max = 2_000
```
RyanJenkinson, line 297: Could we pull this out as a variable in `costs.py` to create 
```
DECOMISSIONING_COSTS_MIN, DECOMISSIONING_COSTS_MAX = 500, 2_000
``` 
as variables and then reference them [here](https://github.com/centrefornetzero/domestic-heating-abm/blob/main/simulation/costs.py#L165) in the function

And then for this test we just `from simulation.costs import DECOMISSIONING_COSTS_MAX`.

Just in case we ever change it so that the test remains robust...

RyanJenkinson (changes requested):
LGTM! Just a very small change before I approve

RyanJenkinson (approved):

PR #102, opened 2022-02-01
Households exclude to-be-banned systems as options, ahead of ban

stephanie-chan-cnz:
- Context: Prior to this PR, households would always ignore the existence of an upcoming ban and react only when the ban is live. This results in a sharp, angular uptake in HPs as of the ban date.
- This PR introduces the possibility of households responding to the ban before it is live - a proportion of households will exclude to-be-banned systems (oil/gas boilers) from their heating system options if the ban has been announced. The proportion of households doing this depends upon the time til the ban. When the ban is live, this is always 100%. When the ban is more than 10 years away, this is always 0%. When the ban is between 0-10 years away, this is determined by a parameterised sigmoid curve (see below). The curve's parameters can be considered placeholder values subject to further research / a future PR.

image

Note: The announcement date can be interpreted specifically as a statement from Government _which leaves no room for uncertainty_. Expected behaviour is that the longer uncertainty persists or a lack of formal ban is put in place (i.e. the less time between announcement and ban dates), the less able heat pump installers are to work at capacity pre-ban and meet the uptick in post-ban demand. For the same ban date, a delay in announcement therefore reduces overall HP uptake.

Screenshot 2022-02-02 at 10 02 02

Screenshot 2022-02-02 at 10 02 15

### DIFF
--- a/simulation/__main__.py
+++ b/simulation/__main__.py
@@ -150,7 +150,7 @@ def parse_args(args=None):
 
     parser.add_argument(
         "--gas-oil-boiler-ban-announce-date",
-        default=datetime.datetime(2025, 1, 1),
+        default=datetime.datetime(2035, 1, 1),
         type=convert_to_datetime,
     )
 
@@ -163,8 +163,17 @@ def parse_args(args=None):
     return parser.parse_args(args)
 
 
+def validate_args(args):
+    if args.gas_oil_boiler_ban_announce_date > args.gas_oil_boiler_ban_date:
+        raise ValueError(
+            f"Boiler ban announcement date must be on or before ban date, got gas_oil_boiler_ban_date:{args.gas_oil_boiler_ban_date}, gas_oil_boiler_ban_announce_date:{args.gas_oil_boiler_ban_announce_date}"
+        )
+
+
 if __name__ == "__main__":
+
     args = parse_args()
+    validate_args(args)
 
     logger.info(
         "parsed arguments",
--- a/simulation/__main__.py
+++ b/simulation/__main__.py
@@ -148,6 +148,12 @@ def parse_args(args=None):
         type=convert_to_datetime,
     )
 
+    parser.add_argument(
+        "--gas-oil-boiler-ban-announce-date",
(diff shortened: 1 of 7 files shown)


simulation/__main__.py
```diff
@@ -148,6 +148,12 @@ def check_string_is_isoformat_datetime(string) -> str:
         type=convert_to_datetime,
     )
 
+    parser.add_argument(
+        "--gas-oil-boiler-ban-announce-date",
+        default=datetime.datetime(2025, 1, 1),
+        type=convert_to_datetime,
```
RyanJenkinson, line 154: I think we also need a check in here that `gas_oil_boiler_ban_announce_date < gas_oil_boiler_ban_date`, perhaps via an assert statement like this somewhere:
```
assert gas_oil_boiler_ban_announce_date < gas_oil_boiler_ban_date, "Ensure the gas oil boiler ban announcement date happens after the ban date"
```

simulation/tests/test_agents.py
```diff
@@ -737,3 +737,93 @@ def test_household_ability_to_choose_heat_pump_as_option_depends_on_model_heat_p
         )
 
         assert all(heat_pump in heating_system_options for heat_pump in HEAT_PUMPS)
+
```
RyanJenkinson, line 740: I think because these are testing how the agents behave under a boiler ban, WDYT about moving them all into their own class:
```
class TestAgentsWithBoilerBan:
```

It might help us segment the desired behaviour under the boiler ban from general behaviour?

simulation/__main__.py
RyanJenkinson: Also when we do this, add a test for this behaviour

RyanJenkinson (changes requested):
LGTM! I think all the logic is good, just some potential comments for how we can simplify a few lines of code (up to you if you accept)

The only change I think we really need is an `assert` statement to make sure that the announcement date is before the ban date (or it would be weird behaviour otherwise, I think) and a corresponding test for this.

RyanJenkinson (approved):

PR #110, opened 2022-02-07
Add sensitivity test jobs

tomwphillips:
Started numbering the sensitivity tests at 50 so there's plenty of room for scenarios.

### DIFF
--- a/k8s/job.jsonnet
+++ b/k8s/job.jsonnet
@@ -49,7 +49,7 @@ local bigquery_arg = [
     '2012-01-01',
     '--steps',
     '120',
-     '--heat-pump-installer-annual-growth-rate',
+    '--heat-pump-installer-annual-growth-rate',
     '0',
   ] + bigquery_arg),
   job('02-%s-baseline' % std.extVar('SHORT_SHA'), [
--- a/k8s/job.jsonnet
+++ b/k8s/job.jsonnet
@@ -40,6 +40,8 @@ local bigquery_arg = [
 ];
 
 [
+  // scenarios
+
   job('01-%s-rhi' % std.extVar('SHORT_SHA'), [
     '--intervention',
     'rhi',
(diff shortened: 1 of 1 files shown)


k8s/job.jsonnet
```diff
@@ -161,5 +162,93 @@ local bigquery_arg = [
...
+  ]),
+
+  job('63-%s-oil-price-low' % std.extVar('SHORT_SHA'), [
+    '--price-gbp-per-kwh-oil',
+    '0.0241',
+  ]),
+  job('64-%s-oil-price-high' % std.extVar('SHORT_SHA'), [
+    '--price-gbp-per-kwh-oil',
+    '0.0723',
+  ]),
```
RyanJenkinson, line 235: Could we change this to only sensitivity test the tuple of values, instead:
```suggestion
  job('59-%s-fuel-price-low' % std.extVar('SHORT_SHA'), [
    '--price-gbp-per-kwh-gas',
    '0.02325',
    '--price-gbp-per-kwh-electricity',
    '0.1003',
    '--price-gbp-per-kwh-oil',
    '0.0241',
  ]),
  job('60-%s-fuel-price-high' % std.extVar('SHORT_SHA'), 
    '--price-gbp-per-kwh-gas',
    '0.06975',
    '--price-gbp-per-kwh-electricity',
    '0.3009',
    '--price-gbp-per-kwh-oil',
    '0.0723',
  ]),
```

RyanJenkinson: If we do this we could renumber all of the scenarios below, but it doesnt matter that much

RyanJenkinson (changes requested):
All the values LGTM, but we can sensitivity test for the tuple of `(gas, elec, oil)` prices, meaning we only have 2 sensitivity tests rather than 6.

PR #120, opened 2022-02-23
Update scenario jobs

RyanJenkinson:
**Note**: Merge this after #118 and #119 

* Adds the new `--include-new-builds` arg in scenario 4 and 5
* Creates a new sub-scenario in Scenario 4 to imagine "unlimited heat pump installer capacity". This will allow us to plot what the heat pump YoY growth _should_ be to meet new+existing building demand in this "best case scenario".

Question: Should this be for scenario 5 instead of scenario 4? Will discuss at stand up, or you can put your thoughts in review.


tomwphillips (dismissed):
@RyanJenkinson LGTM. I've resolved the conflicts with `main`. Please double check that it's correct before you merge.

tomwphillips (approved):

PR #121, opened 2022-02-25
Scale the number of new builds to the agent population size

RyanJenkinson:
Added:
* A scaling factor inside the model.new_builds_per_step property that
  scales according to the model.household_count in relation to the
  ENGLAND_WALES_POPULATION. This means that the number of new builds is
  actually scaled to the number of agents we are running

Changed:
* `ANNUAL_NEW_BUILDS` -> `ENGLAND_WALES_ANNUAL_NEW_BUILDS` so its more
  explicit what the constant represents

Note: This has been tested locally, and is exhibiting the behaviour we would expect (see below) and fixes a bug introduced in #119.
image

### DIFF
--- a/simulation/__main__.py
+++ b/simulation/__main__.py
@@ -12,7 +12,7 @@ import structlog
 from dateutil.relativedelta import relativedelta
 
 from abm import write_jsonlines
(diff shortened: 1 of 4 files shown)


tomwphillips (approved):
LGTM.

PR #136, opened 2024-10-14
Update container

charlotte-avery:
PR required to address deployment error:`Error: The v0 series of google-github-actions/get-gke-credentials is no longer maintained. It will not receive updates, improvements, or security patches.`

### DIFF
--- a/.github/workflows/container.yaml
+++ b/.github/workflows/container.yaml
@@ -71,7 +71,7 @@ jobs:
       - uses: google-github-actions/auth@v0
         with:
           credentials_json: ${{ secrets.DOMESTIC_HEATING_ABM_SA_KEY }}
-      - uses: google-github-actions/get-gke-credentials@v0
+      - uses: google-github-actions/get-gke-credentials@v2
         with:
           cluster_name: cluster
           location: europe-west2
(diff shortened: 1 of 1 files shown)


.github/workflows/container.yaml
```diff
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: google-github-actions/setup-gcloud@v0.2.0
+      - uses: google-github-actions/setup-gcloud@v2
```
shengy90, line 10: I believe you need to auth first before you can setup gcloud with V2:
https://github.com/google-github-actions/setup-gcloud?tab=readme-ov-file#usage

We had this error on Friday where without authenticating, setup-gcloud fails (though GH action allowed you to continue unauthenticated).

shengy90 (changes requested):
Need to auth first before you can setup gcloud I think with V2

.github/workflows/container.yaml
charlotte-avery: Reverted back to v0.2.0

shengy90 (approved):
LGTM

PR #142, opened 2024-10-23
Update model assumptions (EPC upgrade + HP installers)

charlotte-avery:
* Assume that households upgrade to EPC D when they install a HP. 
* Update the initial number of HP installers based on latest data


shengy90 (approved):
lgtm

PR #143, opened 2024-10-24
Increase HP awareness over time

charlotte-avery:
Increase HP awareness over time:
* Initial HP awareness at t=0 defined as input using arg `heat-pump-awareness`
* Add intervention `heat-pump-awareness` which, when applied, increases the number of agents that become HP aware over time
* Add input argument `heat-pump-awareness-intervention-factor` to specify how quickly the heat pump awareness increases over time 
* This factor defines the probability that a given agent, which is not already 'heat pump aware' becomes 'heat pump aware'. Note that the probability increases each year throughout the simulation, and thus the number of 'heat pump unaware' agents decreases year on year in this intervention. 

Update BUS funding cap: 
* 'Standard BUS': £237.5M until 2028. 
--> BUS horizon is 1 Jan 2024 (beginning of simulation) - Apr2025. Assuming £200M Apr 2024-2025 (see [announcement](https://www.gov.uk/government/publications/boiler-upgrade-scheme-approval-to-over-allocate-vouchers/approval-to-over-allocate-vouchers-for-the-boiler-upgrade-scheme-october-2024)) + £37.5M Jan-Apr 2024 given that we assume 2024 had a £150M funding cap. 
* 'Extended BUS': £5.45B until 2035
--> Assuming £250M 2024-’25 + £1.54B 2025-’28 (see [announcement](https://www.gov.uk/government/news/families-business-and-industry-to-get-energy-efficiency-support)) + £515M PA 2028-2035.


simulation/agents.py
```diff
@@ -101,6 +101,12 @@ def reverse_sigmoid(x: float, k: float = SIGMOID_K, offset: float = SIGMOID_OFFS
     return 1 / (1 + math.exp(k * (x + offset)))
 
 
+def sigmoid(x: float, k: float = SIGMOID_K):
+    # Sigmoid function with y bounded between 0 & 1
+
+    return 2 * 1 / (1 + math.exp(-k * (x / 10))) - 1
```
shengy90, line 107: doesn't the formula for sigmoid function take the form 1/(1+exp(-X))? Also what does the denominator '10' here refer to?

shengy90: Is this equation provided by CCC? Or what is the aim we hope to achieve with this equation? At first glance:
- `sigmoid(0)` will give 0
- `sigmoid(9999999) will give ~ 1
- but you need very large `x` value to achieve 1
- looks like `x` =  years, plugging in sigmoid(1) ~ 10%. ,sigmoid(5) ~ 46%
- After 1 (5) year(s) of heat pump awareness campaign only 10% (46%) of people becomes aware of heat pump
- 46% after 5 years feels a bit low, no? Is it realistic?

simulation/costs.py
```diff
@@ -312,7 +312,7 @@ def estimate_boiler_upgrade_scheme_grant(
         return 0
 
     model_population_scale = ENGLAND_WALES_HOUSEHOLD_COUNT_2020 / model.household_count
-    boiler_upgrade_funding_cap_gbp = 450_000_000 / model_population_scale
+    boiler_upgrade_funding_cap_gbp = 250_000_000 / model_population_scale
```
shengy90, line 315: I'm not 100% sure how this change reflects the new announcement + your commit notes:

- The announcement you linked says: 
> The Secretary of State for Energy Security and Net Zero has given permission for Ofgem to over-allocate vouchers by up to £50 million in this financial year, to a total of up to £200 million, to provide certainty that vouchers will continue to be available for the remainder of the financial year.

 which I understood to be:
- total allowable to be allocated: £200M in financial year 1st Apr 2024-31st Mar 2025
- this is £50M above what was originally planned, i.e. previously only £150M allowed over same period

I think Ofgem's website is also saying the same thing as what I understood to be:
https://www.ofgem.gov.uk/environmental-and-social-schemes/boiler-upgrade-scheme-bus

I'm not sure where 250M has come from?

shengy90: If anything my interpretation of total cost of BUS is now:

- 2022 - 2023: £150M
- 2023 - 2024: £150M
- 2024 - 2025: £150M + £50M

Originally was £450M between 2022-2025, i.e. £150M per year.

charlotte-avery: So simulation begins 1 Jan 2024, there is £200M in financial year 1st Apr 2024-31st Mar 2025 so I need to add on an amount 1Jan-1Apr 2024 = £200M / 4 = £50M

simulation/agents.py
charlotte-avery: This is on top of the initial heat pump awareness factor. If, say, the HP awareness factor is 0.25, then 25% of agents already are HP aware. If this function is called, an additional ~50% (of the remaining agents that are not already HP aware) will then become aware of HPs in 5 years time.

shengy90: why couldn't you just use the SIGMOID_K to do this though - as that's also a normalising factor. Also where's the requirement sigmoid(x=10)=0.75 come from?

And also the numerator 2, and -1? Strictly speaking this is no longer sigmoid function too - would suggest renaming to something else which this distribution.

I think this function resemble more like a tanh function instead, not sigmoid?

charlotte-avery: The factor of `2` stretched the function along the `y` and `-1` shifts the y-intercept to 0. This is required to bound the function between 0 & 1 along the y axis

simulation/costs.py
shengy90: Should it then be £150M / 4 though? 

Original grant was £450M over three years i think, i.e. £150M per year, and all the announcement is saying is £50M extra for 2024-2025, bringing total to £500M over 2022-2025?

simulation/agents.py
charlotte-avery: Now using linear function to define probability of becoming HP aware in HP awareness intervention.

simulation/__main__.py
```diff
@@ -146,6 +146,13 @@ def format_uuid(str):
         help="Include new build projections (from constants.py). Installers will also build heat pumps in new builds from 2025.",
     )
 
+    parser.add_argument(
+        "--heat-pump-awareness-intervention-factor",
+        type=float_between_0_and_1,
+        default=0,
+        help="A value between 0 and 1 which determines how quickly heat pump awareness increases over time with the heat pump awareness intervention. A value of 0 indicates no increase in heat pump awareness.",
```
shengy90, line 153: Just to check: is 0 'no increase' or 0 = 'no awareness'?

shengy90 (dismissed):
Approved with 1 comment

simulation/__main__.py
charlotte-avery: No increase in awareness from the initial value. I will clarify in description.

shengy90 (approved):
LGTM

PR #148, opened 2024-11-18
Allow for sudden increase in the heat pump awareness 

charlotte-avery:
Add heat pump awareness campaign intervention:
* Intervention is set up to model a sudden increase in heat pump awareness throughout the population as a result of, for example, some widespread campaign. 
* User inputs the date of the campaign (`heat-pump-awareness-campaign-date`) and the heat pump awareness achieved due to the campaign (`heat-pump-awareness-due-to-campaign`). 
* The heat pump awareness can never decrease from the initial heat pump awareness (`heat-pump-awareness`) even if the user uses a lower value for `heat_pump_awareness_due_to_campaign`. 
* Agents in the population are randomly assigned to be either aware or not aware, before the campaign and after the campaign, with a probability given by `heat-pump-awareness` and  `heat_pump_awareness_due_to_campaign`, respectively. This logic is captured in the functions: `construct_population_awareness` and `construct_population_awareness_due_to_campaign`. 

Code no longer updates to 100% awareness at the boiler ban announcement date:
* Instead, 100% awareness only happens after the ban is in place. 
* If this is effect is desired, one can set the campaign date to the boiler ban announcement date, and set `heat_pump_awareness_due_to_campaign=1`.

### DIFF
--- a/simulation/model.py
+++ b/simulation/model.py
@@ -209,25 +209,55 @@ class DomesticHeatingABM(AgentBasedModel):
         self.households_heat_pump_aware_at_current_step = 0
 
 
-def construct_population_awareness(
+def population_heat_pump_awareness(
     household_population: pd.DataFrame, heat_pump_awareness: float
 ) -> List[bool]:
-    """Construct a list assigning heat pump awareness to each agent in the population"""
-    # Randomly assign heat pump awareness to the agent population with a probability given by `heat_pump_awareness`
+    """Randomly assign heat pump awareness to each agent in the population
+      with a probability given by input param `heat_pump_awareness`
+
+    Args:
+        household_population (pd.DataFrame): household population
+        heat_pump_awareness (float): probability of household becoming heat
+          pump aware
+
+    Returns:
+        List[bool]: list of length equal to the number of households in the
+          population of True/False values depending on whether the household
+            is heat pump aware or not
+    """
     return [
         random.random() < heat_pump_awareness for _ in range(len(household_population))
     ]
 
 
-def construct_population_awareness_due_to_campaign(
-    population_heat_pump_awareness: List[bool],
+def population_heat_pump_awareness_due_to_campaign(
+    agents_heat_pump_awareness: List[bool],
     heat_pump_awareness_due_to_campaign: float,
     heat_pump_awareness: float,
 ) -> List[bool]:
-    """Construct a list assigning heat pump awareness to each agent in the population after the campaign"""
+    """If heat pump campaign is set as intervention in input params to the ABM,
+      then some agents who are not initially heat pump aware will later become
+        heat pump aware at the time of the campaign.
+    This function randomly designates agents to become heat pump aware due to
+      the campaign.
+    The probability of an agent becoming heat pump aware due to the campaign
+      is given by the input param `heat_pump_awareness_due_to_campaign`.
 
-    # Awareness cannot decrease over time
-    # Ensure that the campaign target awareness is at least the current awareness
+    Args:
+        agents_heat_pump_awareness (List[bool]): agent awareness of heat pumps
+          at the beginning of the simulation
+        heat_pump_awareness_due_to_campaign (float): probability of household
+          becoming heat pump aware due to campaign intervention
+        heat_pump_awareness (float): probability of a household being
+          initially heat pump aware irrespecictive the campaign intervention
+
+    Returns:
+        List[bool]: list of length equal to the number of households in the
+          population of True/False values depending on whether the household
+            will become heat pump aware or not due to the campaign intervention
+    """
+    # Ensure that the campaign target awareness is at least the initial awareness
+    # If a user sets a lower target awareness than the initial awareness this is the same as the campaign having no effect on heat pump awareness
     heat_pump_awareness_due_to_campaign = max(
         heat_pump_awareness_due_to_campaign, heat_pump_awareness
     )
@@ -244,17 +274,15 @@ def construct_population_awareness_due_to_campaign(
             heat_pump_awareness_due_to_campaign - heat_pump_awareness
         ) / unaware_population_fraction
 
-    population_heat_pump_awareness_after_campaign = (
-        population_heat_pump_awareness.copy()
-    )
+    agents_heat_pump_awareness_after_campaign = agents_heat_pump_awareness.copy()
     # Randomly assign heat pump awareness to the initially non-aware agent population
-    for i in range(len(population_heat_pump_awareness_after_campaign)):
-        if not population_heat_pump_awareness_after_campaign[i]:
+    for i in range(len(agents_heat_pump_awareness_after_campaign)):
+        if not agents_heat_pump_awareness_after_campaign[i]:
             rand_gen = random.random()
             if rand_gen < target_heat_pump_awareness_for_campaign:
-                population_heat_pump_awareness_after_campaign[i] = True
+                agents_heat_pump_awareness_after_campaign[i] = True
 
-    return population_heat_pump_awareness_after_campaign
+    return agents_heat_pump_awareness_after_campaign
 
 
 def create_household_agents(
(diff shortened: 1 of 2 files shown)


simulation/agents.py
```diff
@@ -686,3 +686,6 @@ def make_decisions(self, model):
             self.heating_system_costs_subsidies = costs_subsidies
             self.heating_system_costs_insulation = costs_insulation
             self.insulation_element_upgrade_costs = chosen_insulation_costs
+
+        if self.is_heat_pump_aware:
+            model.households_heat_pump_aware_at_current_step += 1
```
shengy90, line 706: Is this attribute correct? I'm not seeing this method being declared anywhere...

e.g. `heat_pump_installations_at_current_step` is declared in collectors.py and accessed in model.py

I did see you've created the method  `model_heat_pump_awareness` though.. are you missing some suffix?

simulation/collectors.py
```diff
@@ -277,6 +277,10 @@ def model_heat_pump_installations_at_current_step(model) -> int:
     return model.heat_pump_installations_at_current_step
 
 
+def model_heat_pump_awareness(model) -> int:
+    return model.heat_pump_awareness
```
shengy90, line 285: Is this the right variable? Should you have the `at_current_step` suffix?

simulation/agents.py
```diff
@@ -150,6 +151,9 @@ def __init__(
         self.is_heat_pump_aware = (
             self.heating_system in HEAT_PUMPS or is_heat_pump_aware
         )
+        self.is_heat_pump_aware_after_campaign = (
```
shengy90, line 154: the `self.is_heat_pump_aware` logic is: either (already has heat pump) or (is heat pump aware) because by definition, if they have a heat pump, they must be heat pump aware.

I'm not sure if the same logic follows into `is heat pump aware after campaign`:
- this variable suggests that this household is aware _because_ of the campaign
- not 100% sure 1) if this variable name actually reflects the variable you're trying to create 2) how this variable will be used later on and whether or not it reflects the variable name

will revisit this comment later

simulation/model.py
```diff
@@ -207,13 +209,74 @@ def increment_timestep(self):
...
+    # Randomly assign heat pump awareness to the agent population with a probability given by `heat_pump_awareness`
+    return [
+        random.random() < heat_pump_awareness for _ in range(len(household_population))
+    ]
+
+
+def construct_population_awareness_due_to_campaign(
+    population_heat_pump_awareness: List[bool],
+    heat_pump_awareness_due_to_campaign: float,
+    heat_pump_awareness: float,
```
shengy90, line 225: Just to check:

`heat_pump_awareness_due_to_campaign`: the target awareness we're trying to get to with the campaign
`heat_pump_awareness`: the current level of awareness in the absence of the campaign, i.e. baseline awareness?

This function basically tries to select the people who are currently not yet aware, and randomly assign households s.t. after the campaign the total number of 'aware' household = specified target?

If i'm summarising correctly, please could you:
1. Make sure that the doc string correctly summarises what the function is trying to do, including the argument definitions, as it's not very clear.
2. If my understanding is correct, is `heat_pump_awareness` basically the proportion of `population_heat_pump_awareness` that is `True`? If so - can we clarify this in the code as it's a bit unclear.

simulation/model.py
```diff
@@ -207,13 +209,74 @@ def increment_timestep(self):
...
+) -> List[bool]:
+    """Construct a list assigning heat pump awareness to each agent in the population after the campaign"""
+
+    # Awareness cannot decrease over time
```
shengy90, line 229: This comment and implementation is unclear to me.

The comment `#Awareness cannot decrease over time` suggests that these variables are 'changing' over time but isn't `heat_pump_awareness_due_to_campaign` and input to the simulation (i.e. the target) so it can't really change over time. What changes over time is the `awareness _at_time_step`?

I think the code will be a lot clearer if this logic is elsewhere (possibly right at the beginning of the script where simulation inputs are parsed). Options:
1) or you output a "warning" to in the log to say `heat_pump_awareness_due_to_campaign` value is overriden with `heat_pump_awareness` because value is lower or error or
2) don't let the simulation run and warn the users of invalid simulation input

I think the soft overriding here is very implicit and unclear and I'm trying to guess what you're doing.

simulation/model.py
```diff
@@ -207,13 +209,74 @@ def increment_timestep(self):
...
+    # Ensure that the campaign target awareness is at least the current awareness
+    heat_pump_awareness_due_to_campaign = max(
+        heat_pump_awareness_due_to_campaign, heat_pump_awareness
+    )
+
```
shengy90, line 234: Do we need to go through all these trouble? Why couldn't we simply set `heat_pump_awareness` value to the `heat_pump_awareness_due_to_campaign` value and randomly assign households the target awareness value?

My understanding of `create_household_agents` function where this function `construct_population_awareness_due_to_campaign` is called:

- `create_and_run_simulation` is the handler function
- at the start of simulation, this function will create the `model` (the universe)
- will create households, and the households would have been assigned `heat_pump_aware` at the very beginning
- then `model.run` will collate all the collector functions, and update the households at each timestep

If you want to model heat pump awareness that will increase over time due to the campaign, then shouldn't all this logic live in collector.py, rather than right at the beginning?
-

shengy90: Although I see that in `agents.py` you're only setting the the attribute `is_heat_pump_aware` to this `heat_pump_awareness_due_to_campaign` after the campaign date.

Which suggests that this `heat_pump_awareness_due_to_campaign` is a 'latent' variable that doesn't get activated until after the campaign date.

In which case:
- I'd recommend naming the variable to something clearer e.g. `will_become_heat_pump_aware due to campaign` or something like that or `heat_pump_campaign_target_audience` or something to that effect
- And actually explain this in the doc strings because I'm just trying to infer/ guess what you're trying to do and there's no context elsewhere in the code that actually lets me confirm what I think you're trying to do!

shengy90: And if my guess of what you're trying to implement is correct, I feel like it would be more natural to implement this logic inside `agent.py` instead of creating this value right at the beginning. Creates less attributes/ methods. 

All this logic could have been in `update_heat_pump_awareness` function - just access the `heat_pump_aware_at_current_timestep` value you've created, calculate the gap (target - currently aware), and flip households accordingly to achieve the target.

simulation/agents.py
shengy90: After having read the entire PR, I now understand this variable to be `latent variable` that gets activated later on if the campaign date is reached. But i'm still unsure if we need this variable to implement the campaign effect..

simulation/model.py
charlotte-avery: The reason I have defined awareness in `model.py` r.t. `agents.py` is to ensure that once an agent becomes HP aware, it will always be HP aware. The simulation runs by looping over timesteps for the model horizon, and then at each timestep, loop over all agents (i.e., call functions in `agents.py`). If we change information about agents in `agents.py` in within this for loop, we need to ensure that information is collected and stored somewhere so that at the next timestep, the agent that became 'HP aware' at the previous timestep, is still remembered to be 'HP aware' at the next timestep. In the current ABM implementation, this is achieved by defining whether a household is heat pump aware is defined outside this for loop in `create_household_agents` in `model.py`. Agent properties e.g., EPC, heating system, HP awareness etc. are defined in `create_household_agents` outside the for loop which makes sense to me. 

Given that we can access the full population statistics in `create_household_agents` (we only see one agent at a time in `agents.py`), it makes sense to me to look at who is HP aware, who is not HP here, and then define which of the non-aware HP agents will switch during the campaign. Also, given that 'baseline' or initial HP awareness is defined here in the current ABM implementation, it seems natural to me to also define 'future awareness due to campaign' here.

simulation/collectors.py
charlotte-avery: This is referring to the property `DomesticHeatingABM.heat_pump_awareness` defined in `models.py` which returns `households_heat_pump_aware_at_current_step / household_count`, which is what I want.

simulation/model.py
charlotte-avery: Yes your understanding is correct. I will update the doc strings.

PR #149, opened 2024-11-20
Update household heat pump awareness over time

charlotte-avery:
Add heat pump awareness campaign intervention:

* Intervention is set up to model a sudden increase in heat pump awareness throughout the population as a result of, for example, some widespread campaign.
* User inputs the date of the campaign (`heat-pump-awareness-campaign-date`) and the target heat pump awareness for the campaign (`campaign-target-heat-pump-awareness`).
* If the user inputs `campaign-target-heat-pump-awareness > heat-pump-awareness` a ValueError is raised (campaign should never decreases awareness).
* Agents in the population are randomly assigned to be either aware or not aware on the campaign date with a probability given by `campaign-target-heat-pump-awareness`. 

Code no longer updates to 100% awareness at the boiler ban announcement date:
* Instead, 100% awareness only happens after the ban is in place.
* If this is effect is desired, one can set the campaign date to the boiler ban announcement date, and set `campaign-target-heat-pump-awareness=1`.


simulation/agents.py
```diff
@@ -611,8 +611,32 @@ def compute_heat_pump_capacity_kw(self, heat_pump_type: HeatingSystem) -> int:
...
+    ) -> float:
+        return (
+            model.campaign_target_heat_pump_awareness - model.heat_pump_awareness
+        ) / (1 - model.heat_pump_awareness)
+
+    def update_heat_pump_awareness(self, model) -> None:
+
+        if InterventionType.HEAT_PUMP_CAMPAIGN in model.interventions:
+            if (
+                model.current_datetime == model.heat_pump_awareness_campaign_date
```
shengy90: Is this hard `==` safe? If user input is say '2028-01-31' will this work?

Maybe a safer one would be model.current_datetime >= model.heat_pump_awareness_campaign_date..

shengy90 (dismissed):
Nice - much cleaner and straightforward implementation. Just 1 small comment on the very strict date comparison!

simulation/agents.py
charlotte-avery: I had to change the logging to implement this. See latest commit. 

In summary: we want to say "if the target HP awareness has not been met in the previous timestep (t-1), give a suitable probability of converting agents to becoming HP aware to meet the target at timestep t", however, before the HP awareness was recalculated every timestep (every timestep, HP awareness set to value of zero, then using the line of code in agents.py: `model.households_heat_pump_aware_at_current_step += 1` we count the HP awareness in that timstep). Remember, the simulation loops over each agent every timestep. So if you are on agent num. 3 in the loop, the HP awareness is calculated by looking at how many of the 3 agents are heat pump aware so far, and on timestep 5, we see the HP awareness by  looking at how many of the 5 agents are heat pump aware so far etc. So we need to store the HP awareness at t-1, to access it at t, so we can convert a suitable number of households to become HP aware to meet the target.

simulation/agents.py
```diff
@@ -619,14 +619,14 @@ def proba_of_becoming_heat_pump_aware_required_to_reach_campaign_target(
...
+            model.campaign_target_heat_pump_awareness
+            - model.heat_pump_awareness_at_timestep
+        ) / (1 - model.heat_pump_awareness_at_timestep)
 
     def update_heat_pump_awareness(self, model) -> None:
-
         if (
             InterventionType.HEAT_PUMP_CAMPAIGN in model.interventions
-            and model.current_datetime == model.heat_pump_awareness_campaign_date
+            and model.current_datetime >= model.heat_pump_awareness_campaign_date
```
shengy90, line 629: How confident are you with this implementation? Is it worth having a test to test this logic? Up to you to decide!

shengy90 (approved):
LGTM, just a suggest for test but leave it up to you to decide!

shengy90 (approved):
👍

shengy90 (approved):

PR #151, opened 2024-12-02
Treat HP campaign args same as HP discount args

charlotte-avery:
Run ABM with HP awareness starting at 25%, increasing to 50% at 2030 and 75% by 2034. To achieve this, allow for user to run ABM with multiple inputs for the HP awareness campaign similar to the inputs for HP discount args. 

Update grant caps for BUS.


simulation/__main__.py
```diff
@@ -199,10 +195,22 @@ def validate_args(args):
             f"Boiler ban announcement date must be on or before ban date, got gas_oil_boiler_ban_date:{args.gas_oil_boiler_ban_date}, gas_oil_boiler_ban_announce_date:{args.gas_oil_boiler_ban_announce_date}"
         )
 
-    if args.campaign_target_heat_pump_awareness < args.heat_pump_awareness:
-        raise ValueError(
-            f"Campaign target awareness must be greater than or equal to the population heat pump awareness, got campaign_target_heat_pump_awareness:{args.campaign_target_heat_pump_awareness}, heat_pump_awareness:{args.heat_pump_awareness}"
+    if args.campaign_target_heat_pump_awareness_date is not None:
+        # Check that target awareness inputs increase over the model horizon
+        campaigns = sorted(args.campaign_target_heat_pump_awareness_date)
```
shengy90, line 200: This should be a function that we unit test I think!

k8s/job.jsonnet
```diff
@@ -45,10 +45,32 @@ local job(name, args_excl_output) = {
     'extended_boiler_upgrade_scheme',
     '--intervention',
     'heat_pump_campaign',
-    '--campaign-target-heat-pump-awareness',
-    '0.8',
-    '--heat-pump-awareness-campaign-date',
-    '2028-01-01',
+    '--campaign-target-heat-pump-awareness-date',
```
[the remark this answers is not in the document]
shengy90: You can do it as a JSON string that gets loaded as a dictionary:
https://stackoverflow.com/questions/18006161/how-to-pass-dictionary-as-command-line-argument-to-python-script

simulation/__main__.py
charlotte-avery: See `test_campaign_target_less_than_heat_pump_awareness_raises_value_error` in `test_main`

shengy90: Oh I mean the logic inside this `if` statement - it's a pretty convoluted logic that I think we should write as a function, then test that given inputs you get expected output.

e.g.

```python
     if args.campaign_target_heat_pump_awareness_date is not None:
           x = parse_campaign_dates_and_target(arg1, arg2)

     def test_parse_campaign_dates_and_target():
           test this function.
```

Increase legibility of the code, and if someone changes some stuff then this test will break, and they'd have to either update the tests (if break is intended) or fix the change to conform to the test etc.

k8s/job.jsonnet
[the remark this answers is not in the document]
charlotte-avery: Based on our discussion I will keep the current methodology.

shengy90 (approved):
Approving but should test `check_parsed_target_heat_pump_awareness` explicitly and with parametrised tests as well next time!

